PR #1, opened 2017-10-13
Add coveralls

afalko:
@justinharringa

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,7 @@
 
     <properties>
         <slf4j.version>1.7.25</slf4j.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -91,6 +92,31 @@
...
+                    <execution>
+                        <id>post-unit-test</id>
+                        <goals>
+                            <goal>report</goal>
+                        </goals>
+                        <phase>test</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
```
justinharringa: This could go into pluginManagement instead of the main dependencies.

PR #3, opened 2017-10-23
Initial fixes for integration tests

afalko:


dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
```diff
@@ -194,20 +192,24 @@ public void updateStore(String store, String img, String tag) throws IOException
...
         }
-        updateStoreOnGithub(storeRepo, STORE_JSON_FILE, img, tag);
+        updateStoreOnGithub(storeRepo, Constants.STORE_JSON_FILE, img, tag);
     }
 
     protected void updateStoreOnGithub(GHRepository repo, String path, String img, String tag) throws IOException {
-        GHContent content;
         try {
-            content = repo.getFileContent(path);
+            repo.getFileContent(path);
```
justinharringa: Is there a reason we need to to this once without the `latestCommit` and then do it again with `latestCommit`?

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/CommandLineTest.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Created by afalko on 10/25/17.
+ */
+public class CommandLineTest {
+    @Test
+    public void testLoadSubcommands() throws Exception {
+        Set<ClassPath.ClassInfo> classes = CommandLine.findSubcommands(CommandLine.getArgumentParser());
+        assertEquals(classes.size(), 3);
```
justinharringa: why `3`?

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtilTest.java
```diff
@@ -41,7 +41,7 @@
     public void testGetGithubUtil() throws Exception {
         githubUtil = Mockito.mock(GithubUtil.class);
         dockerfileGithubUtil = new DockerfileGithubUtil(githubUtil);
-        assertEquals(dockerfileGithubUtil.getGithubUtil(), githubUtil);
+        Assert.assertEquals(dockerfileGithubUtil.getGithubUtil(), githubUtil);
```
justinharringa: Why add this in? You're statically importing the class.

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Test
+    public void testParseStoreToImagesMap() throws Exception {
+        DockerfileGithubUtil dockerfileGithubUtil = mock(DockerfileGithubUtil.class);
+        when(dockerfileGithubUtil.getMyself()).thenReturn(mock(GHMyself.class));
+        when(dockerfileGithubUtil.getRepo(anyString())).thenReturn(mock(GHRepository.class));
+        GHContent mockContent = mock(GHContent.class);
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+        when(mockContent.read()).thenReturn(classloader.getResourceAsStream("image-store-sample.json"));
```
justinharringa: Any reason this couldn't be inline instead of jumping to files?

dockerfile-image-update-itest/src/main/java/com/salesforce/dockerfileimageupdate/itest/TestStarter.java
```diff
@@ -0,0 +1,95 @@
...
+import org.testng.xml.XmlSuite;
+
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Created by afalko, smulakala, minho.park on 7/14/16.
+ */
+public class TestStarter {
```
justinharringa: Is there an advantage of using this rather than just running something like [Maven Failsafe](http://maven.apache.org/surefire/maven-failsafe-plugin/)

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
afalko: The logic here essentially is "if file contents are missing, then create the store file in the catch block". I would love it if the lib allowed me to do this with an if statement, but unfortunately that is not the case.

afalko: Also, I was hitting up a nasty behavior with the APIs, where if I didn't get the content with the latest commit, it would return the first commit. If there is no file yet, then most likely there is no commit in the repo and we'd get NPEs later.

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/CommandLineTest.java
afalko: Thats the number of subcommands that we currently have.

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
afalko: I felt this would be more readable and there wouldn't be any json embedded in a string or something like that.

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtilTest.java
afalko: This is a mistake on my part; not sure how I fat fingered this in...will revert.

dockerfile-image-update-itest/src/main/java/com/salesforce/dockerfileimageupdate/itest/TestStarter.java
afalko: We didn't want to exec this with maven...we run it standalone in docker-compose.

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/CommandLineTest.java
afalko: See new update, made this clearer and improved the assert

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
justinharringa: The things I generally don't like about this approach are:
* You get a filesystem involved
* You have to switch contexts to figure out what's in the file rather than stay within the same test

dockerfile-image-update-itest/src/main/java/com/salesforce/dockerfileimageupdate/itest/TestStarter.java
justinharringa: Can't that kick off Maven?

dockerfile-image-update-cmdtool/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
afalko: Feel free to send a PR to improve it ;)

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
```diff
@@ -117,7 +115,7 @@ public void modifyAllOnGithub(GHRepository repo, String branch,
 
     protected void modifyOnGithubRecursive(GHRepository repo, GHContent content,
                                            String branch, String img, String tag) throws IOException {
-        if (content.getType().equals(GITHUB_FILE)) {
+        if (content.getType().equals(Constants.GITHUB_FILE)) {
```
justinharringa: Use `content.isFile()`

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
```diff
@@ -194,20 +192,24 @@ public void updateStore(String store, String img, String tag) throws IOException
...
+            repo.createContent("", "initializing store", path);
+
         }
-        if (content.getType().equals(GITHUB_FILE)) {
+
+        String latestCommit = repo.getBranches().get(repo.getDefaultBranch()).getSHA1();
+        log.info("Loading image store at commit {}", latestCommit);
+        GHContent content = repo.getFileContent(path, latestCommit);
+
+        if (content.getType().equals(Constants.GITHUB_FILE)) {
```
justinharringa: Use `content.isFile()`. Once done, no need for `GITHUB_FILE`

dockerfile-image-update-itest/src/main/java/com/salesforce/dockerfileimageupdate/itest/TestStarter.java
afalko: Perhaps, but experimenting with that would be for another PR

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
afalko: Nice, this must be a new feature in with the library version bump. Fixed.

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
afalko: See comment above

dockerfile-image-update-cmdtool/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGithubUtil.java
justinharringa: Seems like we might be able to use a `GHSearchBuilder` to do this in a less `catch`-y way. 😆  If you don't mind, it'd probably be nice to have a comment there stating the purpose and that another type of API call may be necessary.

justinharringa: Something like https://github.com/username/image-tag-store-repo/search?utf8=%E2%9C%93&q=filename%3Astore.json+path%3A%2F&type= may do the trick

PR #5, opened 2018-01-22
Prep for Maven Central pushing

justinharringa:


.travis.yml
```diff
@@ -3,19 +3,24 @@ env:
...
 - docker-compose up
 after_success:
 - mvn coveralls:report
-- if [ "${TRAVIS_BRANCH}" == "master" ]; then 
-      docker login -u="${DOCKER_USER}" -p="${DOCKER_PASSWORD}";
-      docker push salesforce/dockerfile-image-update; 
-  fi
+deploy:
+  provider: script
+  script: "./travis.deploy.sh"
```
afalko: `.travis`, not ./travis

.travis.deploy.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+cp .travis.settings.xml $HOME/.m2/settings.xml
```
afalko: `set -e`
${HOME}

pom.xml
```diff
@@ -12,14 +12,47 @@
...
+    <name>${project.groupId}:${project.artifactId}</name>
+    <description>Parent pom ${project.groupId}:${project.artifactId}.</description>
+    <url>https://github.com/salesforce/dockerfile-image-update</url>
+    <organization>
+        <name>Salesforce</name>
+        <url>http://www.salesforce.com</url>
+    </organization>
+    <licenses>
+        <license>
+            <name>BSD 3-Clause</name>
```
afalko: BSD 3-Clause License

pom.xml
```diff
@@ -12,14 +12,47 @@
...
+        <developer>
+            <name>Andrey Falko</name>
+            <email>afalko@salesforce.com</email>
+            <organization>Salesforce</organization>
+        </developer>
+        <developer>
+            <name>Justin Harringa</name>
+            <email>jharringa@salesforce.com</email>
+            <organization>Salesforce</organization>
+        </developer>
```
afalko: Add Minho to developers?

pom.xml
```diff
@@ -119,4 +158,117 @@
...
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+    <distributionManagement>
```
afalko: I don't normally see projects have this in their pom.xmls; even those that push to maven central; perhaps they all push manually?

justinharringa: Generally, if you're pushing to Maven Central this is either in the POM or the settings.xml

pom.xml
justinharringa: Done in 1f009e9d69b838eca260dd6e89810f4c47d98f1a

.travis.deploy.sh
justinharringa: ${HOME} is in 90c7042438b8236c961cb14028b620139bbbc830
I believe Travis is going to do `set -e` for us in the [script deployment](https://docs.travis-ci.com/user/deployment/script/) which is a more specialized use case than scripts above. If they don't end up doing that, then I'll add it.

.travis.yml
justinharringa: Fixed in d8fb072657a729c83c472065c3861347f5528fd9

pom.xml
justinharringa: Here's an example from the [multi-module maven plugin](https://github.com/danielflower/multi-module-maven-release-plugin/blob/master/pom.xml)

PR #7, opened 2018-01-22
Change perms on .travis.deploy.sh, add `set -ex`

justinharringa:


justinharringa:
@afalko ALL THE CODE COVERAGE! 😄

PR #9, opened 2018-01-23
Abort on container exit for docker-compose

justinharringa:


justinharringa:
@afalko fixed the integration test jar location

PR #12, opened 2018-01-24
Developer connection correction (JGit / insteadOf)

justinharringa:
I suspect that the git config's [url.<base>.insteadOf](https://www.kernel.org/pub/software/scm/git/docs/git-config.html#_variables) could be tripping us up here. Let's check it out and resolve this...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ script:
 - rm itest.env
 - mkdir -p ${HOME}/.ssh
 - mv -f id_rsa_dockerfile_image_update ${HOME}/.ssh/id_rsa
-- git config -l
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -19,6 +19,7 @@ script:
 - rm itest.env
 - mkdir -p ${HOME}/.ssh
 - mv -f id_rsa_dockerfile_image_update ${HOME}/.ssh/id_rsa
+- git config -l
```
afalko: Is this for debugging?

justinharringa: Yep. Removed it in c5db720. Should be good to go @afalko

PR #14, opened 2018-01-27
Add encrypted gpg key for code signing

justinharringa:


justinharringa:
@afalko This should be the final step necessary to close this out unless there is some complication in decryption / import of the gpg signing key. Thanks for bearing with the churn.

justinharringa:
The previous iteration was activating the right profile but I erroneously assumed GPG signing was being handled for us (duh... of course you need a key... argh...). The successful deploy only got the artifacts staged but javadoc & sources weren't being sent, artifacts weren't signed and Nexus also wasn't being told to auto-close.

.travis.settings.xml
```diff
@@ -15,6 +15,7 @@
             </activation>
             <properties>
                 <gpg.executable>gpg</gpg.executable>
+                <gpg.keyname>${env.GPG_KEY_NAME}</gpg.keyname>
```
afalko: Can someone make a PR and have mvn echo this?

justinharringa:
Nope. Travis hides it all. This just uses environment variable resolution.
Also the passphrase is only for signing as this is a subkey. 😄

justinharringa:
Also, Nexus staging will validate whether this has been revoked if we find
that this has happened. Just FYI

PR #16, opened 2018-02-22
Release shaded jar - Let shade plugin overwrite versioned artifact

justinharringa:
This is default behavior and we likely wouldn't use the unshaded jar for
anything. Also need to add dependency-reduced-pom.xml to .gitignore

### DIFF
--- a/dockerfile-image-update/pom.xml
+++ b/dockerfile-image-update/pom.xml
@@ -79,7 +79,6 @@
                                     <mainClass>com.salesforce.dockerfileimageupdate.CommandLine</mainClass>
                                 </transformer>
                             </transformers>
-                            <finalName>dockerfile-image-update</finalName>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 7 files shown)


dockerfile-image-update/pom.xml
```diff
@@ -79,7 +79,6 @@
                                     <mainClass>com.salesforce.dockerfileimageupdate.CommandLine</mainClass>
                                 </transformer>
                             </transformers>
-                            <finalName>dockerfile-image-update</finalName>
```
afalko: Don't we need to update Dockerfile?

justinharringa: Yeah, done in b7d4c94 but also needed to resolve the fact that the default artifact is shaded so itest no longer needs dependencies on GSON and GitHub API.

afalko: And what about https://github.com/salesforce/dockerfile-image-update/blob/master/README.md#running-locally ?

Is there a maven copy target you can do to preserve the "nice" final name?

PR #17, opened 2018-02-22
Don't build tags/branches

justinharringa:
Tags and branches starting with dockerfile-image-update should not be
built

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ sudo: required
 language: java
 services:
 - docker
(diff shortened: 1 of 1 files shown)


justinharringa:
To test this, I created a branch starting with dockerfile-image-update- and the result was that the request to build stopped:
image

afalko:
Is this to make travis not build tags that get added?

justinharringa:
Yep

PR #20, opened 2018-03-07
Do not attempt to fork forks

justinharringa:
This causes errors in the way that we do this. This change will make
that unsupported until we invest the time to handle that properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,10 @@ In environment variables, please provide:
  * `git_api_url` : This is the Endpoint URL of the GitHub API. In general GitHub, this is `https://api.github.com/`; for Enterprise, this should be `https://hostname/api/v3`. (this variable is optional; you can provide it through the command line.)
 
 ### Precautions
-This tool may create a LOT of forks in your account. All pull requests created are through a fork on your own account.
+1. This tool may create a LOT of forks in your account. All pull requests created are through a fork on your own account.
+2. We currently do not operate on forked repositories due to limitations in forking a fork on GitHub.
+We should invest some time in doing this right. 
+
 
 ### How to use it
 Our recommendation is to run it as a docker container:
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ In environment variables, please provide:
 ### Precautions
 1. This tool may create a LOT of forks in your account. All pull requests created are through a fork on your own account.
 2. We currently do not operate on forked repositories due to limitations in forking a fork on GitHub.
-We should invest some time in doing this right. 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -54,7 +54,10 @@ In environment variables, please provide:
  * `git_api_url` : This is the Endpoint URL of the GitHub API. In general GitHub, this is `https://api.github.com/`; for Enterprise, this should be `https://hostname/api/v3`. (this variable is optional; you can provide it through the command line.)
 
 ### Precautions
-This tool may create a LOT of forks in your account. All pull requests created are through a fork on your own account.
+1. This tool may create a LOT of forks in your account. All pull requests created are through a fork on your own account.
+2. We currently do not operate on forked repositories due to limitations in forking a fork on GitHub.
+We should invest some time in doing this right.
```
afalko: "We have the following issue open to resolve this", then create an issue and link to it.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -113,9 +116,14 @@ protected void forkRepositoriesFound(Map<String, String> parentToPath,
              * in the map above, find the list of repositories under the authorized user, and iterate through that list.
              */
             GHRepository parent = c.getOwner();
-            log.info("Forking {}...", parent.getFullName());
-            parentToPath.put(c.getOwner().getFullName(), c.getPath());
-            dockerfileGitHubUtil.checkFromParentAndFork(parent);
+            if (parent.isFork()) {
+                log.info("Skipping {} because it's a fork already. Sending a PR to a fork is unsupported at the moment.",
```
afalko: Should be a `warn`, no?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -153,10 +153,13 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
      * OkUrlFactory, which is required to specify the cache. In other words, we cannot flush the cache.
      *
      * Instead, we wait for 60 seconds if the list retrieved is not the list we want.
+     *
+     *
+     * TODO: Did updating the GitHub api fix this?
```
afalko: Is this actionable? Feels like this TODO will be there forever :)

justinharringa: Fair 😄

PR #27, opened 2018-07-21
Make comment more effective

afalko:
I chose to leave the logic alone and only make the comment and logging clearer. The new version works as designed with submodules: https://df17-demo.net/jenkins/job/afalko/job/alpine-python/job/master/


afalko:
NP @vishmegh88, I wanted to get you unblocked ASAP. No reason to have you waiting on nit-picky changes :)

vishmegh88:
Thanks @afalko :)

jdoshi1:
@afalko - Would be good to add some text in [README.md#precautions](https://github.com/salesforce/dockerfile-image-update#precautions) that we don't support updating Dockerfiles that are part of submodules.

PR #37, opened 2018-08-15
Fixed additional edge cases with trailing comments and whitespace

afalko:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -18,6 +19,8 @@
 import java.nio.file.Paths;
 import java.util.List;
 
+import static java.lang.Character.SPACE_SEPARATOR;
```
brian-mcnamara: Is this used?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -8,6 +8,7 @@
 
 package com.salesforce.dockerfileimageupdate.utils;
 
+import com.google.common.base.CharMatcher;
```
brian-mcnamara: This too looks unused.

afalko:
Thanks @brian-mcnamara ! I had my ide refactor imports and it found a few more.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java
```diff
@@ -404,25 +404,36 @@ public void testFindImagesAndFix_doNotDeleteOtherLines() throws Exception {
         assertEquals(strB.toString(), "hello\nFROM image:newtag\nthis is a test\n\n\n\nworld\n");
     }
 
-    @Test
-    public void testFindImagesAndFix_doNotDeletePostTagData() throws Exception {
+    @DataProvider
+    public Object[][] postTagData() {
```
justinharringa: Wouldn't it be better to do this kind of testing directly against `getEndOfTagLength(String)` to keep the locus of test smaller and narrow scope (smaller unit for the unit test)?

afalko:
@justinharringa I found additional edge cases that we were missing and added a separate test for the length finder. I fixed three bugs with this exercise: 
1. Tag would get wiped out if tab followed comment.
2. You can now have a comment with a tag: `image:tag # :commented-out-tag`
3. Empty tag returns the right thing

PR #38, opened 2018-09-04
Switch to alpine

afalko:
This speeds up download of the tool for demos


justinharringa:
@afalko looks like we have to get your CLA signature now

justinharringa:
🚢 🚀

PR #47, opened 2019-01-29
Log if skipping an archived repo with Dockerfiles

justinharringa:
Just adding a simple diagnostic if there is an archived repo that we're skipping.


justinharringa:
This code definitely needs a refactor. Most of `changeDockerfiles` in `Parent` and `All` is duplication. It also looks like there may be an issue where we only update the first image from a parent in the forked repo.

PR #48, opened 2019-05-03
Fix to avoid indefinite loop in case of failure to fork a repo

VijayKumarMidde:
Currently, if we fail to fork a repo we are just quietly catching the exception and not propagating the exception up. Because of this, before updating Dockerfiles and sending PRs, we may go into an indefinite loop waiting for the forking to complete. This PR contains fix to such cases.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -95,7 +95,7 @@ public class All implements ExecutableWithNamespace {
                                          PagedSearchIterable<GHContent> contentsWithImage,
                                          String image) throws IOException {
         log.info("Forking {} repositories...", contentsWithImage.getTotalCount());
-        List<String> parentReposForked = new ArrayList<>();
+        List<String> parentReposAlreadyChecked = new ArrayList<>();
         GHRepository parent;
         String parentRepoName = null;
         for (GHContent c : contentsWithImage) {
@@ -115,10 +115,14 @@ public class All implements ExecutableWithNamespace {
                 pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
                 imagesFoundInParentRepo.put(parentRepoName, image);
 
-                // fork the parent if not already forked
-                if (!parentReposForked.contains(parentRepoName)) {
-                    dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    parentReposForked.add(parentRepoName);
+                // fork the parent if not already forked or we couldn't fork
+                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                    GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
+                    }
+                    parentReposAlreadyChecked.add(parentRepoName);
                 }
             }
         }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
@@ -37,6 +37,11 @@ public class Child implements ExecutableWithNamespace {
         log.info("Retrieving repository and creating fork...");
         GHRepository repo = dockerfileGitHubUtil.getRepo(ns.get(Constants.GIT_REPO));
         GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(repo);
+        if (fork == null) {
+            log.info("Unable to fork {}. Please make sure that the repo is forkable.",
+                    repo.getFullName());
+            return;
+        }
 
         if (branch == null) {
             branch = repo.getDefaultBranch();
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
@@ -110,7 +110,7 @@ public class Parent implements ExecutableWithNamespace {
     protected Multimap<String, String> forkRepositoriesFoundAndGetPathToDockerfiles(PagedSearchIterable<GHContent> contentsWithImage) throws IOException {
         log.info("Forking repositories...");
         Multimap<String, String> pathToDockerfilesInParentRepo = HashMultimap.create();
-        List<String> parentReposForked = new ArrayList<>();
+        List<String> parentReposAlreadyChecked = new ArrayList<>();
         GHRepository parent;
         String parentRepoName = null;
         for (GHContent c : contentsWithImage) {
@@ -128,11 +128,15 @@ public class Parent implements ExecutableWithNamespace {
                         parentRepoName);
             } else {
                 pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                // fork the parent if not already forked
-                if (!parentReposForked.contains(parentRepoName)) {
+                // fork the parent if not already forked or we couldn't fork
+                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
                     log.info("Forking {}", parentRepoName);
-                    dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    parentReposForked.add(parentRepoName);
+                    GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
+                    }
+                    parentReposAlreadyChecked.add(parentRepoName);
                 }
             }
         }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -64,7 +64,7 @@ public class GitHubUtil {
     /* Create a fork. Does not check if the fork is already there, because the fork will not be created if a fork
      * already exists.
      */
-    public GHRepository createFork(GHRepository repo) throws IOException {
+    public GHRepository createFork(GHRepository repo) {
         try {
             return repo.fork();
         } catch (IOException e) {
@@ -169,6 +169,7 @@ public class GitHubUtil {
                 String forkName = s.substring(s.lastIndexOf('/') + 1);
                 log.info(forkName);
                 if (!repoNamesSet.contains(forkName)) {
+                    log.debug("Forking is still in progress for {}" , forkName);
                     listOfReposHasRecentForks = false;
                 }
             }
--- a/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ChildTest.java
+++ b/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ChildTest.java
@@ -16,6 +16,7 @@ import org.mockito.Mockito;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
+import java.io.IOException;
 import java.util.Collections;
 import java.util.Map;
 
@@ -51,6 +52,7 @@ public class ChildTest {
         Namespace ns = new Namespace(inputMap);
         DockerfileGitHubUtil dockerfileGitHubUtil = Mockito.mock(DockerfileGitHubUtil.class);
         Mockito.when(dockerfileGitHubUtil.getRepo(Mockito.any())).thenReturn(new GHRepository());
+        Mockito.when(dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(Mockito.any())).thenReturn(new GHRepository());
         doNothing().when(dockerfileGitHubUtil).modifyAllOnGithub(Mockito.any(), Mockito.any(), Mockito.any(), Mockito.any());
         doNothing().when(dockerfileGitHubUtil).updateStore(anyString(), anyString(), anyString());
         doNothing().when(dockerfileGitHubUtil).createPullReq(Mockito.any(), anyString(), Mockito.any(), anyString());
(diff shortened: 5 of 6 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/exception/UnableToForkException.java
```diff
@@ -0,0 +1,14 @@
+package com.salesforce.dockerfileimageupdate.exception;
+
+
+/**
+ * Indicates that there was an error while trying to fork a Git repository.
+ */
+public class UnableToForkException extends Exception {
```
justinharringa: This seems unnecessary. Does the GitHub library not send a typed exception for this?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
```diff
@@ -66,8 +67,8 @@ public void testCreateFork() throws Exception {
         verify(repo, times(1)).fork();
     }
 
-    @Test
-    public void testCreateForkWithExceptionProceeds() throws Exception {
+    @Test(expectedExceptions = UnableToForkException.class)
```
justinharringa: Probably should be it's own new test since other exceptions should probably proceed.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/exception/UnableToForkException.java
VijayKumarMidde: Github library always throws IOException. I thought it would be more readable this way.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
VijayKumarMidde: Other exceptions can also cause indefinite loop. So for any exception thrown during forking we should propagate it up.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/exception/UnableToForkException.java
VijayKumarMidde: Github library always throws IOException.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
VijayKumarMidde: Actually it should not proceed if it fails to create a fork. Otherwise we would run into indefinite loop again.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -68,9 +68,8 @@ public GHRepository createFork(GHRepository repo) throws IOException {
         try {
             return repo.fork();
         } catch (IOException e) {
-            log.error("Could not fork {}", repo.getFullName(), e);
+            throw new IOException(String.format("Unable to fork %s", repo.getFullName()), e);
         }
-        return null;
```
justinharringa: Was anything using this? This seems like a behavioral change.

Is this an exceptional case?

VijayKumarMidde: Returned value is not used anywhere.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -117,7 +117,16 @@ protected void forkRepositoriesFound(Multimap<String, String> pathToDockerfilesI
 
                 // fork the parent if not already forked
                 if (!parentReposForked.contains(parentRepoName)) {
-                    dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    try {
```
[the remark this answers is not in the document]
VijayKumarMidde: We can't extract this directly into `DockerfileGitHubUtil`. If we did, then it would work for `all` & `parent` subcommands but doesn't work `child` subcommand. 
But what we could extract this into another class `X`, where `X` inherits from `ExecutableWithNamespace` and, class `All` & class `Parent` would inherit from `X`.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
```diff
@@ -36,7 +36,14 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
         log.info("Retrieving repository and creating fork...");
         GHRepository repo = dockerfileGitHubUtil.getRepo(ns.get(Constants.GIT_REPO));
-        GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(repo);
+        GHRepository fork;
+        try {
+             fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(repo);
+        } catch (IOException e) {
+            log.info("Could not fork {}", repo.getFullName(), e);
+            String msg = String.format("Unable to fork %s. Please make sure that the repo is forkable.");
+            throw new IllegalAccessException(msg);
```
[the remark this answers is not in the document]
VijayKumarMidde: Yup. Added a test case.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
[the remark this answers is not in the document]
justinharringa: I don't think inheritance would be necessary in this scenario.

justinharringa:
In your description, you mention `Currently, if we fail to fork a repo we are just quietly catching the exception and not propagating the exception up. Because of this, before updating Dockerfiles and sending PRs, we may go into an indefinite loop waiting for the forking to complete. This PR contains fix to such cases.` 

If I understand the problem correctly, the issue is that if there is an error when forking, we log the error and continue but that fork was added to a list of expected repos so we wait indefinitely in [GitHubUtil.getGHRepositories](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java#L158) (which is called out as a workaround... probably something we should fix but not something you would need to address here). 

Is the case that you're trying to solve for when people [disallow forking private repos](https://help.github.com/en/articles/allowing-people-to-fork-private-repositories-in-your-organization)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
justinharringa: Any reason you couldn't just check to see if the return value is null from this rather than using exceptions for flow control?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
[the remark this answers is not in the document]
justinharringa: We actually don't know if the `IOException` was exactly the case you're solving for if I understand correctly so switching to that exception doesn't seem correct.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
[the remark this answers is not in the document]
justinharringa: Seems like you could use the previous implementation of `closeOutdatedPullRequestAndFork` and just check to see if it returned a fork (if it's null, it was not successful and presumably logged - you could remove from the `pathToDockerfilesInParentRepo` multimap and you would not want to add it to the `parentReposForked` list).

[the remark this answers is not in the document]
VijayKumarMidde: In this case we would still have duplicate code.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
VijayKumarMidde: Modified the flow to check for return value.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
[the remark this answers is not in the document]
VijayKumarMidde: Library throws IOException if any failure happens when forking. Anyways now I updated this to check for return value.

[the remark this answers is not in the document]
justinharringa: Why are we throwing an `IllegalAccessException` if it may not be related to this use case? It could be a network error or other things as well.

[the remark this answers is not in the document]
justinharringa: There may be a good reason, I'm just asking BTW :)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -131,7 +131,11 @@ protected void loadDockerfileGithubUtil(DockerfileGitHubUtil _dockerfileGitHubUt
                 // fork the parent if not already forked
                 if (!parentReposForked.contains(parentRepoName)) {
                     log.info("Forking {}", parentRepoName);
-                    dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
+                    }
                     parentReposForked.add(parentRepoName);
```
justinharringa: Should we be adding to `parentReposForked` if this repo was not forked?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -117,7 +117,11 @@ protected void forkRepositoriesFound(Multimap<String, String> pathToDockerfilesI
 
                 // fork the parent if not already forked
                 if (!parentReposForked.contains(parentRepoName)) {
-                    dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
+                    }
                     parentReposForked.add(parentRepoName);
```
justinharringa: Should we be adding to `parentReposForked` if this repo was not forked?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -169,6 +169,7 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
                 String forkName = s.substring(s.lastIndexOf('/') + 1);
                 log.info(forkName);
                 if (!repoNamesSet.contains(forkName)) {
+                    log.info("Forking is still in progress for {}" , forkName);
```
justinharringa: Will this get chatty in the logs? Would it maybe be better for `debug` mode? A screenshot of output in the PR comment might be helpful to see what it looks like.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: `parentReposForked` is not used outside of this method. It will avoid trying to create fork and failing multiple times if we are updating multiple Dockerfiles in a same fork.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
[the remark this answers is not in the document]
VijayKumarMidde: In this case, we need to stop the execution flow and return if we are unable to create a PR. Would you recommend to throw any other exception?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
VijayKumarMidde: parentReposForked is not used outside of this method. It will avoid trying to create fork and failing multiple times if we are updating multiple Dockerfiles in a same fork.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
VijayKumarMidde: This won't chatty in the logs because we have 60 sec sleep. This log is important for us to know if the execution has gone into infinite loop again.
Screen Shot 2019-05-06 at 7 49 16 PM

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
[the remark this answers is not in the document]
justinharringa: Is there a reason you couldn't simply log a message and return instead of throwing an exception? I guess I don't see a need to introduce another exception when you can handle flow control without it.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
justinharringa: Might be nice to rename this since we may not have forked all of the repos in this list. Possibly `parentReposForked`->`parentReposAlreadyChecked`. Either that, or change the comment above https://github.com/salesforce/dockerfile-image-update/pull/48/files#diff-da5f0732ed5f695522bdefddf18a2555R132 to say something like `fork the parent if not already forked or we couldn't fork`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
VijayKumarMidde: Updated PR.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
VijayKumarMidde: Updated PR

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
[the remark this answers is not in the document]
VijayKumarMidde: Makes sense. Updated PR.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
justinharringa: Does a user care about this especially since there are already a lot of log messages? I was wondering if you could send a screenshot of a bunch of log messages to see how this looks in the flow.

justinharringa:
LGTM. Thanks for the contribution and for answering the questions I had! Appreciate it!

PR #49, opened 2019-05-09
Update dockerhub setup

justinharringa:
The previous creds were expired for the Docker Hub security incident. New cred time.


.travis.deploy.sh
```diff
@@ -18,5 +18,5 @@ cp .travis.settings.xml ${HOME}/.m2/settings.xml
 mvn releaser:release
 
 # Push latest docker image
-docker login -u="${DOCKER_USER}" -p="${DOCKER_PASSWORD}"
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
npwolf: For safety, do you want to wrap this in a set +x echo set -x . Guessing travis probably hides this anyway, but just a thought.

afalko: why remove curlies?

justinharringa: Secrets are hidden in Travis

justinharringa: I'll add curlies for curliness without running`curl` :smile:

justinharringa: I'll wrap it cuz why not

justinharringa: `curly` `+x-x` FTW

PR #52, opened 2019-05-16
Avoid infinite sleep loop if multiple fork names come back from GitHub API

VijayKumarMidde:
When we try to fork a repo and fail for some reason, we were still adding that repo to [parentReposAlreadyChecked](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L113) as it would help us to avoid trying to create a fork and failing multiple times, when we are updating multiple Dockerfiles in the same repository. But this is not completely correct as we add the `(repo, dockerfile)` first to [pathToDockerfilesInParentRepo](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L130) and then we check. Consider the following case:

If [contentsWithImage](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L116) is `['seanhuitest/helloworld-docker-child', 'vmidde/helloworld-docker-child', 'vmidde/helloworld-docker-child',]`. In this case we have two Dockerfiles in `vmidde/helloworld-docker-child` repository which needs to be updated. Also assume that `vmidde/helloworld-docker-child` is unforkable.

In this case, when we first [visit `vmidde/helloworld-docker-child` while iterating](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L116) we add and remove `vmidde/helloworld-docker-child` from [pathToDockerfilesInParentRepo](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L112). In the same iteration we also add `vmidde/helloworld-docker-child` to [parentReposAlreadyChecked](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L139). When we [visit `vmidde/helloworld-docker-child` for the second time,](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L116) we first add `vmidde/helloworld-docker-child` to [pathToDockerfilesInParentRepo](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L130) and we never remove it because it is already checked in the previous iteration.

To fix this issue, I'm simply not adding a repo to [parentReposAlreadyChecked](https://github.com/salesforce/dockerfile-image-update/blob/362eab5d527131f6d588fb7a4ce36abd176c4fae/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java#L139) if we are unable to fork the repo.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -113,15 +113,15 @@ public class All implements ExecutableWithNamespace {
                         parentRepoName);
             } else {
                 // fork the parent if not already forked
-                if (!parentReposForked.contains(parentRepoName)) {
+                if (parentReposForked.contains(parentRepoName) == false) {
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork != null) {
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                    } else {
                         // Add repos to pathToDockerfilesInParentRepo and imagesFoundInParentRepo only if we forked it successfully.
                         pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
                         imagesFoundInParentRepo.put(parentRepoName, image);
                         parentReposForked.add(parentRepoName);
-                    } else {
-                        log.info("Could not fork {}", parentRepoName);
                     }
                 }
             }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -95,7 +95,7 @@ public class All implements ExecutableWithNamespace {
                                          PagedSearchIterable<GHContent> contentsWithImage,
                                          String image) throws IOException {
         log.info("Forking {} repositories...", contentsWithImage.getTotalCount());
-        List<String> parentReposAlreadyChecked = new ArrayList<>();
+        List<String> parentReposForked = new ArrayList<>();
         GHRepository parent;
         String parentRepoName = null;
         for (GHContent c : contentsWithImage) {
@@ -112,17 +112,17 @@ public class All implements ExecutableWithNamespace {
                 log.warn("Skipping {} because it's a fork already. Sending a PR to a fork is unsupported at the moment.",
                         parentRepoName);
             } else {
-                pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                imagesFoundInParentRepo.put(parentRepoName, image);
-
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork == null) {
+                    if (fork != null) {
+                        // Add repos to pathToDockerfilesInParentRepo and imagesFoundInParentRepo only if we forked it successfully.
+                        pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
+                        imagesFoundInParentRepo.put(parentRepoName, image);
+                        parentReposForked.add(parentRepoName);
+                    } else {
                         log.info("Could not fork {}", parentRepoName);
-                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
                     }
-                    parentReposAlreadyChecked.add(parentRepoName);
                 }
             }
         }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
@@ -128,15 +128,15 @@ public class Parent implements ExecutableWithNamespace {
                         parentRepoName);
             } else {
                 // fork the parent if not already forked
-                if (!parentReposForked.contains(parentRepoName)) {
+                if (parentReposForked.contains(parentRepoName) == false) {
                     log.info("Forking {}", parentRepoName);
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork != null) {
+                    if (fork == null) {
+                        log.info("Could not fork {}", parentRepoName);
+                    } else {
                         // Add repos to pathToDockerfilesInParentRepo only if we forked it successfully.
                         pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
                         parentReposForked.add(parentRepoName);
-                    } else {
-                        log.info("Could not fork {}", parentRepoName);
                     }
                 }
             }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
@@ -110,7 +110,7 @@ public class Parent implements ExecutableWithNamespace {
     protected Multimap<String, String> forkRepositoriesFoundAndGetPathToDockerfiles(PagedSearchIterable<GHContent> contentsWithImage) throws IOException {
         log.info("Forking repositories...");
         Multimap<String, String> pathToDockerfilesInParentRepo = HashMultimap.create();
-        List<String> parentReposAlreadyChecked = new ArrayList<>();
+        List<String> parentReposForked = new ArrayList<>();
         GHRepository parent;
         String parentRepoName = null;
         for (GHContent c : contentsWithImage) {
@@ -127,16 +127,17 @@ public class Parent implements ExecutableWithNamespace {
                 log.warn("Skipping {} because it's a fork already. Sending a PR to a fork is unsupported at the moment.",
                         parentRepoName);
             } else {
-                pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
                     log.info("Forking {}", parentRepoName);
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork == null) {
+                    if (fork != null) {
+                        // Add repos to pathToDockerfilesInParentRepo only if we forked it successfully.
+                        pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
+                        parentReposForked.add(parentRepoName);
+                    } else {
                         log.info("Could not fork {}", parentRepoName);
-                        pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
                     }
-                    parentReposAlreadyChecked.add(parentRepoName);
                 }
             }
         }
--- a/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
+++ b/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
@@ -25,10 +25,9 @@ import java.util.ArrayList;
 import java.util.Map;
 import java.util.Set;
 
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.anyString;
-import static org.mockito.Matchers.eq;
+import static org.mockito.Matchers.*;
 import static org.mockito.Mockito.*;
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
 
 /**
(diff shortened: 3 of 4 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -115,14 +115,16 @@ protected void forkRepositoriesFound(Multimap<String, String> pathToDockerfilesI
...
 
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
                     if (fork == null) {
                         log.info("Could not fork {}", parentRepoName);
                         pathToDockerfilesInParentRepo.remove(parentRepoName, c.getPath());
+                        imagesFoundInParentRepo.remove(parentRepoName, image);
```
justinharringa: This seems to be what really affects the logic. It seems like you reverted the name as well as the comment here https://github.com/salesforce/dockerfile-image-update/pull/52/files#diff-850febb926ac48d58dc0962f42249092R118

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
```diff
@@ -61,7 +60,53 @@ public void testForkRepositoriesFound() throws Exception {
         all.loadDockerfileGithubUtil(dockerfileGitHubUtil);
         all.forkRepositoriesFound(ArrayListMultimap.create(), ArrayListMultimap.create(), contentsWithImage, "image");
 
-        Mockito.verify(dockerfileGitHubUtil, times(1)).closeOutdatedPullRequestAndFork(any());
+        Mockito.verify(dockerfileGitHubUtil, times(3)).closeOutdatedPullRequestAndFork(any());
+    }
+
+    @Test
+    public void testForkRepositoriesFound_unableToforkRepo() throws Exception {
```
justinharringa: Please either add a comment on what this is testing. It seems like it would take some time to figure out what went wrong with this test if it started failing.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
```diff
@@ -109,6 +109,49 @@ public void testForkRepositoriesFound() throws Exception {
         assertEquals(repoMap.size(), 3);
     }
 
+    @Test
+    public void testForkRepositoriesFound_unableToforkRepo() throws Exception {
```
justinharringa: Same. What are we testing? What is the expected result here?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
VijayKumarMidde: Added comments.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
VijayKumarMidde: Added comments.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: Are you saying with this fix, the logic is being affected?

justinharringa: @VijayKumarMidde I don't think renaming `parentReposAlreadyChecked` to `parentReposForked` or putting `parentReposForked.add(parentRepoName)` in the else has any effect. We also lost part of the comment: https://github.com/salesforce/dockerfile-image-update/pull/52/files#diff-850febb926ac48d58dc0962f42249092L118

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
```diff
@@ -109,6 +109,55 @@ public void testForkRepositoriesFound() throws Exception {
...
+        /**
+         * Suppose we have multiple dockerfiles that need to updated in a repo and we fail to fork such repo,
+         * we should not add those repos to pathToDockerfilesInParentRepo.
+         */
+        DockerfileGitHubUtil dockerfileGitHubUtil = mock(DockerfileGitHubUtil.class);
+
+        GHRepository contentRepo1 = mock(GHRepository.class);
+        when(contentRepo1.getFullName()).thenReturn("1");
+
+        GHRepository contentRepo2 = mock(GHRepository.class);
```
justinharringa: Better name for this? `duplicateContentRepo1` maybe?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: @justinharringa If you don't mind, can you please elaborate why you think putting `parentReposForked.add(parentRepoName)` in the else doesn't have any effect? Thanks!

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
VijayKumarMidde: Sounds good. I will update.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
justinharringa: It's not returned and doesn't modify anything that is passed in. What effect did it have moving it into the `else`?

VijayKumarMidde: Before this change, we used to add repo name to `parentReposForked ` always (ie., even if the fork was not created). 

Say, our `contentsWithImage` array looks like ['abc/Dockerfile', 'abc/Dockerfile'] (if Github returns duplicates) and also assume that repo `abc` is unforkable. In this case, we first add `('abc', 'Dockerfile')` to `pathToDockerfilesInParentRepo` and we fail to fork `abc` repo. So we remove `('abc', 'Dockerfile')` from `pathToDockerfilesInParentRepo`. But we are still adding `abc` to `parentReposForked` set. In the next iteration of for loop we again add `('abc', 'Dockerfile')` to `pathToDockerfilesInParentRepo` and see that `abc` is present in `parentReposForked`. This time we DONT remove `('abc', 'Dockerfile')` from `pathToDockerfilesInParentRepo`. So it goes into infinite loop again.

To fix the above case, I'm adding repo name to `parentReposForked ` only in else condition (ie., only when a fork is created successfully).

VijayKumarMidde: I should probably add a comment in the code why I'm doing this. let me know if this is clear. I will add some comments to code and update the PR.

VijayKumarMidde: Or I could simply add to `pathToDockerfilesInParentRepo` and `imagesFoundInParentRepo` only if `fork` is not `null`.

justinharringa: @VijayKumarMidde yeah, I'm in favor of simplifying the logic here and getting rid of the `remove()`s. Does `forkRepositoriesFoundAndGetPathToDockerfiles_unableToforkRepo` cover the case where GitHub returns the same name twice? You might put that in the name.

Comments would be super helpful. 

Thanks and sorry about the delay on this.

afalko: Since `imagesFoundInParentRepo` is passed into this function, it makes me nervous that we are mutating it. If the problem is that dupes are showing up, then could we change this to another form of `Map` that had the fork path as the key?

VijayKumarMidde:
> Awaiting simplification of logic (see comments)

Updated PR.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: @afalko `imagesFoundInParentRepo` is passed as an empty `Map` to this function. This function was updating `imagesFoundInParentRepo` before these changes.

I think we can't use another form of Map with fork path as the Key because, in some cases we may need to update multiple Dockerfiles present in a repo.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -127,16 +127,17 @@ protected void loadDockerfileGithubUtil(DockerfileGitHubUtil _dockerfileGitHubUt
...
             } else {
-                pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
                     log.info("Forking {}", parentRepoName);
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork == null) {
+                    if (fork != null) {
```
justinharringa: Could you flip this to avoid negative conditionals?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -112,17 +112,17 @@ protected void forkRepositoriesFound(Multimap<String, String> pathToDockerfilesI
...
-                pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                imagesFoundInParentRepo.put(parentRepoName, image);
-
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
                     GHRepository fork = dockerfileGitHubUtil.closeOutdatedPullRequestAndFork(parent);
-                    if (fork == null) {
+                    if (fork != null) {
```
justinharringa: Could you flip this to avoid negative conditionals?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -112,17 +112,17 @@ protected void forkRepositoriesFound(Multimap<String, String> pathToDockerfilesI
                 log.warn("Skipping {} because it's a fork already. Sending a PR to a fork is unsupported at the moment.",
                         parentRepoName);
             } else {
-                pathToDockerfilesInParentRepo.put(parentRepoName, c.getPath());
-                imagesFoundInParentRepo.put(parentRepoName, image);
-
-                // fork the parent if not already forked or we couldn't fork
-                if (!parentReposAlreadyChecked.contains(parentRepoName)) {
+                // fork the parent if not already forked
+                if (!parentReposForked.contains(parentRepoName)) {
```
justinharringa: Could you make this a positive assertion? 
```suggestion
                if (parentReposForked.contains(parentRepoName) == false) {
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
justinharringa: @afalko seems like some of this should be refactored to remove duplication and move more towards immutable calls. I could see some of these things being pulled into first-class objects rather than passing around a bunch of `MultiMap`s but that might be a tall order for this PR? Definitely open to ideas though.

justinharringa: (part of what I meant to say was that the mutation of the `MultiMap` is something that has been here for a while and wasn't introduced by Vijay)

afalko: Yeh, maybe file an issue for that and move forward here

justinharringa: I opened #71

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
VijayKumarMidde: @justinharringa Updated PR.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: @justinharringa Updated PR.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
VijayKumarMidde: @justinharringa Updated PR.

PR #73, opened 2019-06-08
Improve context in diagnostic logs

justinharringa:
Will add context for each change. Overall, there were some confusing strings and it seems useful to log some of these things near their locus of error.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -72,13 +72,13 @@ public class All implements ExecutableWithNamespace {
                 changeDockerfiles(ns, pathToDockerfilesInParentRepo, imagesFoundInParentRepo, imageToTagMap, currUserRepo,
                         skippedRepos);
             } catch (IOException e) {
+                log.error(String.format("Error changing Dockerfile for %s", currUserRepo.getName()), e);
                 exceptions.add(e);
             }
         }
 
         if (!exceptions.isEmpty()) {
-            log.info("There were {} errors with changing Dockerfiles.", exceptions.size());
-            throw exceptions.get(0);
+            throw new IOException(String.format("There were %s errors with changing Dockerfiles.", exceptions.size()));
         }
 
         if (!skippedRepos.isEmpty()) {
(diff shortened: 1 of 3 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -72,13 +72,13 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                 changeDockerfiles(ns, pathToDockerfilesInParentRepo, imagesFoundInParentRepo, imageToTagMap, currUserRepo,
                         skippedRepos);
             } catch (IOException e) {
+                log.error(String.format("Error changing Dockerfile for %s", currUserRepo.getName()), e);
```
justinharringa: Seems advantageous to go ahead and log this immediately to be able to envision flow a bit.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -72,13 +72,13 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
                 exceptions.add(e);
             }
         }
 
         if (!exceptions.isEmpty()) {
-            log.info("There were {} errors with changing Dockerfiles.", exceptions.size());
-            throw exceptions.get(0);
+            throw new IOException(String.format("There were %s errors with changing Dockerfiles.", exceptions.size()));
```
justinharringa: I'm not stuck on this approach but it seemed odd to log the message and then just throw the first exception.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -142,7 +142,7 @@ protected void loadDockerfileGithubUtil(DockerfileGitHubUtil _dockerfileGitHubUt
             }
         }
 
-        log.info("Path to Dockerfiles in repo '{}': {}", parentRepoName, pathToDockerfilesInParentRepo);
+        log.info("Path to Dockerfiles in repos: {}", pathToDockerfilesInParentRepo);
```
justinharringa: `parentRepoName` was the last repo (which doesn't really make sense when you're printing something for all of the repos.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -167,7 +167,7 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
             boolean listOfReposHasRecentForks = true;
             for (String s : pathToDockerfileInParentRepo.keySet()) {
                 String forkName = s.substring(s.lastIndexOf('/') + 1);
-                log.info(forkName);
+                log.info(String.format("Verifying that %s has been forked", forkName));
```
justinharringa: No context

justinharringa:
@afalko any objections?

PR #78, opened 2019-07-02
Fix for updater missing sending PRs for few repos

VijayKumarMidde:
Sometimes updater is missing to send PRs to all the repos. To get the list of repositories updater uses `GHMyself.listRepositories()` method that returns a `PagedIterable<GHRepository>` closure. We loop through this `Iterable` to verify if all the forks were created and loop through the same `Iterable` again to send PRs. Sometimes if Git is flaky, `Iterable` fails to return all the repos in the second iteration and updater misses sending PRs to those repos.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -181,8 +181,8 @@ public class GitHubUtil {
     }
 
     /**
-     * Returns a <code>java.util.Map</code> of GitHub repositories owned by a user. Returned map contains repository names
-     * mapped to their corresponding GitHub repository objects.
+     * Returns a <code>java.util.Map</code> of GitHub repositories owned by a user. Returned Map's keys are the repository
+     * names and values are their corresponding GitHub repository objects.
      * @param user
      * @return
      */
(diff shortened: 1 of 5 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -180,7 +177,24 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
                 Thread.sleep(TimeUnit.MINUTES.toMillis(1));
             }
         }
-
         return listOfRepos;
     }
+
+    /**
+     * Returns a <code>java.util.Map</code> of GitHub repositories owned by a user. Returned map contains repository names
```
justinharringa: Could perhaps say that the Map's key is the name of the repo.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -180,7 +177,24 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
...
+     * mapped to their corresponding GitHub repository objects.
+     * @param user
+     * @return
+     */
+    public Map<String, GHRepository> getReposForUserAtCurrentInstant(GHMyself user) {
+        Map<String, GHRepository> repoByName = new HashMap<>();
+        if (user == null) {
+            return repoByName;
```
justinharringa: Might be useful to cover this with a unit test. Was this situation covered by a test for `getGHRepositories()`?

VijayKumarMidde: Added unit test for the `null` case.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
VijayKumarMidde: Updated PR

PR #131, opened 2020-03-31
Use JDK/JRE 8 as before

justinharringa:
The full CI didn't lock to a consistent JDK/JRE until recent GitHub
Actions work. When Travis moved to JDK 11, it brought sads when
generating javadoc. Instead of taking on a migration to a newer JDK and
introducing more modules stuff, locking to JDK/JRE 8 as before.

Improvements to this:
* Use a JDK_VERSION in Makefile to drive all container JDK/JRE
dependencies
* Put releaser:release on batch/quiet mode (could potentially just use
batch)
* Correct some Javadoc
* Tell Maven Javadoc plugin that we have level 8 source
* Check that all env vars are set in .ci.deploy.sh


justinharringa:
Javadoc bug: https://bugs.openjdk.java.net/browse/JDK-8212233

PR #137, opened 2020-04-02
Use versioned jar in itests

justinharringa:


justinharringa:
Closing in favor of changes in #138

PR #138, opened 2020-04-02
Use versioned jar in itests and test suite starter improvements

justinharringa:


justinharringa:
This is an improvement which would close #137

PR #142, opened 2020-04-17
Don't close forks, don't close PRs for child/parent commands

### LINKED ISSUE #72: Stop closing forks / PRs - instead look at refreshing fork
As mentioned in #32 we probably should not be closing out existing PRs / forks because if there are multiple images that a repository depends on, it should receive at least one PR for all updates or a PR for each update. 

The impetus for closing them was to handle situations where files had moved or there were potentially other conflicts. (See [comment](https://github.com/salesforce/dockerfile-image-update/blob/424d4cc/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L43))

Could simply create new PRs based off of a new branch and leave around old ones or try and resolve conflicts within (which seems like a more difficult task).

justinharringa:
It has proven wasteful and confusing to remove forks.

This changes the `parent` and `child` commands to use branches with a name based on the new `image-tag` combination.

This will have the following benefits:
* Stop closing forks and creating new ones
  * This will keep PRs for other image-tag combos open
* Create a branch specific to the image-tag combo we're updating to (which provides more context for the PR)
  * A benefit here is that it is also easier to locate DFIU PRs without requiring the usage of a specific PR body ([PULL_REQ_ID](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java#L30) will no longer be necessary and we can provide more context in the PR body).

Parent and Child are covered by #72 - All will keep the current strategy.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
@@ -1,0 +1,86 @@
+package com.salesforce.dockerfileimageupdate.model;
+
+// Converts an imageName to a branchName. Primary conversion necessary is : to -
+// Also support backward compatible method of specifying a branch
+//
+// Container name rules: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+// Branch Rules: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html
+//
+public class GitForkBranch {
+    private final String branchPrefix;
+    private final String imageName;
+    private final String imageTag;
+    private final boolean specifiedBranchOverride;
+
+    public GitForkBranch(String imageName, String imageTag) {
+        this(imageName, imageTag, null);
+    }
+
+    public GitForkBranch(String imageName, String imageTag, String specifiedBranch) {
+        this.imageTag = (imageTag == null || imageTag.trim().isEmpty()) ? "" : imageTag.trim();
+        this.imageName = (imageName == null || imageName.trim().isEmpty()) ? "" : imageName.trim();
+        if (specifiedBranch == null || specifiedBranch.trim().isEmpty()) {
+            if (this.imageName.isEmpty()) {
+                throw new IllegalArgumentException("You must specify an imageName if not specifying a branch");
+            }
+            this.branchPrefix = this.imageName.replace(":", "-").toLowerCase();
+            this.specifiedBranchOverride = false;
+        } else {
+            this.branchPrefix = specifiedBranch;
+            this.specifiedBranchOverride = true;
+        }
+    }
+
+    /**
+     * If a specifiedBranch was specified, do a direct comparison. Otherwise, check for a
+     * translated imageName as a prefix.
+     *
+     * @param branchName branch name to check against our new image
+     * @return is the branch name equal to specifiedBranch or the normalized image name
+     */
+    public boolean isSameBranchOrHasImageNamePrefix(String branchName) {
+        if (this.imageTag == null) {
+            return this.getBranchName().equals(branchName);
+        } else if (branchName != null) {
+            String tempBranchName = branchName.trim();
+            return getBranchWithoutTag(tempBranchName).equals(this.branchPrefix);
+        }
+        return false;
+    }
+
+    private String getBranchWithoutTag(String branchName) {
+        int lastDash = branchName.lastIndexOf("-");
+        if (lastDash == -1) {
+            return branchName;
+        } else {
+            return branchName.substring(0, lastDash);
+        }
+    }
+
+    /**
+     * @return a branch was specified to override default behavior
+     */
+    public boolean useSpecifiedBranchOverride() {
+        return this.specifiedBranchOverride;
+    }
+
+    /**
+     * Either the specified branch of a combo of normalized imageName-imageTag
+     * @return
+     */
+    public String getBranchName() {
+        if (this.imageTag == null || this.imageTag.trim().isEmpty()) {
+            return this.branchPrefix;
+        } else {
+            return this.branchPrefix + "-" + this.imageTag;
+        }
+    }
+
+    public String getImageName() {
+        return imageName;
+    }
+
+    public String getImageTag() {
+        return imageTag;
+    }
+}
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
@@ -1,20 +1,17 @@
 package com.salesforce.dockerfileimageupdate.model;
 
-// Converts an imageName to a branchName. Primary conversion necessary is : to -
-// Also support backward compatible method of specifying a branch
-//
-// Container name rules: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
-// Branch Rules: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html
-//
+/**
+ * Converts an imageName to a branchName. Primary conversion necessary is : to -
+ * Also support backward compatible method of specifying a branch
+ *
+ * Container name rules: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+ * Branch Rules: https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html
+ */
 public class GitForkBranch {
     private final String branchPrefix;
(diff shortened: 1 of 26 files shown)


justinharringa:
Note: There's likely a need for fork cleanup that should either be a part of this process or separate (to reduce specific time of this process). For example, the original parent repo could be deleted or moved which would put the fork in the bot org in a weird state (the deletion can make the fork in the bot org a non-fork).

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
```diff
@@ -0,0 +1,86 @@
...
+        if (this.imageTag.equals("")) {
+            return this.getBranchName().equals(branchName);
+        } else if (branchName != null) {
+            String tempBranchName = branchName.trim();
+            return getBranchWithoutTag(tempBranchName).equals(this.branchPrefix);
+        }
+        return false;
+    }
+
+    private String getBranchWithoutTag(String branchName) {
```
afalko, line 51: Are we documenting branch name conventions anywhere?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -0,0 +1,91 @@
...
+            try (InputStream stream = content.read(); InputStreamReader streamR = new InputStreamReader(stream)) {
+                try {
+                    json = JsonParser.parseReader(streamR);
+                } catch (JsonParseException e) {
+                    log.warn("Not a JSON format store. Clearing and rewriting as JSON...");
+                    json = JsonNull.INSTANCE;
+                }
+            }
+            String jsonOutput = getAndModifyJsonString(json, img, tag);
+            content.update(jsonOutput,
```
afalko, line 67: Doe this include retries?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -205,79 +201,19 @@ protected boolean changeIfDockerfileBaseImageLine(String imageToFind, String tag
...
     public void createPullReq(GHRepository origRepo,
                               String branch, GHRepository forkRepo,
-                              String message) throws InterruptedException, IOException {
-        if (message == null) {
-            message = "Automatic Dockerfile Image Updater";
+                              String title) throws InterruptedException, IOException {
+        if (title == null) {
+            title = "Automatic Dockerfile Image Updater";
         }
+        // TODO: This may loop forever in the event of constant -1 pullRequestExitCodes...
```
afalko, line 214: Give it a configurable timeout?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -287,23 +223,76 @@ public void createPullReq(GHRepository origRepo,
...
+     */
+    public Optional<PagedSearchIterable<GHContent>> getGHContents(String org, String img)
+            throws IOException, InterruptedException {
+        PagedSearchIterable<GHContent> contentsWithImage = null;
+        for (int i = 0; i < 5; i++) {
+            contentsWithImage = findFilesWithImage(img, org);
+            if (contentsWithImage.getTotalCount() > 0) {
+                break;
+            } else {
+                Thread.sleep(1000);
```
afalko, line 287: `TimeUnit.SECONDS.sleep(1);`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -72,12 +75,18 @@ public GHRepository createFork(GHRepository repo) {
...
+     */
     public void safeDeleteRepo(GHRepository repo) throws IOException {
-        try {
-            repo.delete();
-        } catch (IOException e) {
-            throw new IOException("Please verify that the GitHub token " +
-                    "provided has access to deleting repositories.");
+        if (!repo.queryPullRequests().state(GHIssueState.OPEN).list().iterator().hasNext()) {
+            try {
+                repo.delete();
```
afalko, line 85: Need to retry on failure a few times?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
justinharringa: Yeah, right now focusing on solidifying the move to stop forking and branch instead. This needs to be done at some point or another though.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
justinharringa: In which scenario would you expect that? Are you thinking of a network blip or momentary server unavailability? A potential issue with that would be to potentially put pressure on GitHub rate limits.

One thing that I'm considering to open up as an issue is to avoid forking at all if a repo doesn't need a change. We should be able to detect that off of the primary repo and then just skip forking it at the beginning.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
justinharringa: Nah, it doesn't currently. I haven't observed many issues where retries set this in a bad spot. That is something that could be considered for sure. I would have a similar concern about determining when to retry to help reduce rate limit pressure.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/model/GitForkBranch.java
justinharringa: Yeah, I kinda put it up top but this should be nearer to the function. Adding.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
afalko: Yeh, network blips or errors

justinharringa:
This PR already involves a fairly significant amount of change. Talked a bit with @afalko. Will let `parent` and `child` get these new changes and then `all` will kind of operate as it did before (though not a complete fork delete). Next PR will involve completely aligning `all` with `parent`.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -212,7 +120,24 @@ protected void changeDockerfiles(Namespace ns,
...
         if (isContentModified) {
-            dockerfileGitHubUtil.createPullReq(parent, branch, forkedRepo, ns.get(Constants.GIT_PR_TITLE));
+            // TODO: get the new PR number and cross post over to old ones
+            dockerfileGitHubUtil.createPullReq(parent, gitForkBranch.getBranchName(), forkedRepo, ns.get(Constants.GIT_PR_TITLE));
+            // TODO: Run through PRs in fork to see if they have head branches that match the prefix and close those?
         }
     }
+    private Optional<GHPullRequest> commentAndCloseExistingPullRequestsForBranch(GHRepository parentRepo, GitForkBranch gitForkBranch) {
+        Optional<GHPullRequest> specifiedBranchPullRequest = dockerfileGitHubUtil.getPullRequestForImageBranch(parentRepo, gitForkBranch);
+        //            GHPullRequest pullRequest = specifiedBranchPullRequest.get();
```
afalko, line 130: Need to keep this commented?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
afalko: Change still not made?

justinharringa: Bah... There were a bunch... Corrected that crust

Issue #149, opened 2020-05-18
PR body should be descriptive instead of UUID


justinharringa:
Once we migrate the `all` command over to using branches instead of the UUID, we can tee a PR off with a more descriptive body. MVP seems to be a specific body based on image and tag.

PR #152, opened 2020-05-19
Add a meaningful PR body for parent/child commands

justinharringa:
This will add a body like the following to `parent` and `child` commands instead of `f9ed6ea5-6e74-4338-a629-50c5c6807a6b`:

> `alpine` changed recently. This pull request ensures you're using the latest version of the image and changes `alpine` to the latest tag: `42`
>
> New value: `alpine:42`

After this, we just need to do this for the `all` subcommand in order for it to also handle #149


justinharringa:
`createPullReq` should get some refactoring when commonality is brought into the `all` command.

justinharringa:
@afalko do you mind poking this one again? The update from `master` invalidated your review.

image

PR #161, opened 2020-06-09
Add GitHub CodeQL analysis

justinharringa:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 16 * * 6'
+
+jobs:
+  CodeQL-Build:
+
+    # CodeQL runs on ubuntu-latest and windows-latest
```
afalko, line 12: Why not alpine?

PR #163, opened 2020-06-13
Additional wait on fork creation before creating branch

justinharringa:
This helps in a scenario where we haven't given GitHub / GHE the
requisite "5 seconds" to create its fork. We'll wait 10 seconds for an
extra buffer if we don't see that the parent's default branch is not in the fork.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
  */
 public class GitHubUtil {
     private static final Logger log = LoggerFactory.getLogger(GitHubUtil.class);
+    public static final String NO_BRANCH_WARN_FORMAT = "Couldn't find branch `%s` in repo `%s`. Waiting a second...";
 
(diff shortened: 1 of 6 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -184,12 +212,32 @@ public GHContent tryRetrievingContent(GHRepository repo, String path, String bra
                 break;
             } else {
                 log.info("Waiting for GitHub API cache to clear...");
-                Thread.sleep(TimeUnit.MINUTES.toMillis(1));
```
afalko, line 187: Why not just do `TimeUnit.MNUTES.sleep()`?

justinharringa: The extraction of `waitFor` is mostly for improvements in unit tests (you don't actually have to wait in the tests). I had considered `waitFor(seconds)` but thought that kind of bucks normal standards of millis and allows the caller to be somewhat explicit with `TimeUnit.${UNIT}.toMillis(${NUM})`. I also considered `waitForMinutes()`, `waitForSeconds()`, etc... but then thought I was going too far and this seemed to express things.

PR #192, opened 2021-01-26
Fix potential NPE

stephenc:
A noddy NPE that my IDE was complaining about


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -123,8 +123,10 @@ public void modifyAllOnGithub(GHRepository repo, String branch,
                 getGitHubUtil().waitFor(TimeUnit.SECONDS.toMillis(1));
             }
         }
-        for (GHContent con : tree) {
-            modifyOnGithubRecursive(repo, con, branch, img, tag);
+        if (tree != null) {
```
afalko, line 126: Put this check before the break above, that way we retry and error if this stays null

afalko (changes requested):
Move null check to retry loop

afalko (approved):

justinharringa (approved):

PR #193, opened 2021-01-26
Add support for extra body content in PRs

stephenc:
Adds the ability to add some extra text to the PR body.

Use case: we are using a separate bot account to run the pull requests, so we need visibility into the PRs that get created in order to ensure that they get some attention. Also for the case where the teams managing the repos against which the PRs are created, we need the ability to point them at where the PRs originate from so that they have tracability on who to blame if the PRs become a nusance


afalko (approved):
This is a great addition. Thank you!

justinharringa:
Agree with @afalko! Thanks for the contribution @stephenc!

PR #194, opened 2021-01-26
Add excludes regex based on repository name

stephenc:
Use case: 

We have some repositories that build collections of base images. We want those repos to always have `:latest` as the tag as they are building the chain of potential updates


afalko (commented):
Overall, this looks good; it looks like it is failing due to a slight drop in code coverage.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/process/GitHubPullRequestSender.java
```diff
@@ -53,6 +56,11 @@ public GitHubPullRequestSender(DockerfileGitHubUtil dockerfileGitHubUtil, Forkab
             totalContentsFound++;
             parent = ghContent.getOwner();
             parentRepoName = parent.getFullName();
+            if (excludes != null && excludes.matcher(parent.getName()).matches()) {
+                log.info("Skipping repository {} as it name ({}) matches the excludes regex",
```
justinharringa, line 60:
```suggestion
                log.info("Skipping repository {} as its name ({}) matches the excludes regex",
```

justinharringa:
Labeler fails because of a fork but it's fine and it's time to merge. Sorry for the reeeeeeaally long time to respond here @stephenc  Related issue: https://github.com/actions/labeler/issues/12

justinharringa (approved):

PR #211, opened 2021-03-23
Fix issue 210: https://github.com/salesforce/dockerfile-image-update/…

ksrisurapaneni:
…issues/210

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
@@ -80,6 +80,15 @@ public class DockerfileGitHubUtil {
         if (query.substring(query.lastIndexOf(' ') + 1).length() <= 1) {
             throw new IOException("Invalid image name.");
         }
+        // When the query string is a combination of dashes and dots, github API search doesn't work properly and doesn't find the image
+        // ex: xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime
+        // https://github.com/craftcms/cms/issues/3080
+        // To fix this, we need to replace the dashes with spaces in the search string and Github search can then find the image
+        // New string: xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime
+        // It doesn't effect other functions and works perfectly
+        if(query.contains("-")) {
+            query = query.replace('-', ' ');
+        }
         search.q("\"FROM " + query + "\"");
         log.debug("Searching for {}", query);
         PagedSearchIterable<GHContent> files = search.list();
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
@@ -87,7 +87,7 @@ public class DockerfileGitHubUtil {
         // New string: xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime
         // It doesn't effect other functions and works perfectly
         if(query.contains("-")) {
-            query = query.replaceAll('-', ' ');
+            query = query.replace('-', ' ');
         }
         search.q("\"FROM " + query + "\"");
         log.debug("Searching for {}", query);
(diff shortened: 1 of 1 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -80,6 +80,15 @@ public GHRepository getRepo(String repoName) throws IOException {
...
             throw new IOException("Invalid image name.");
         }
+        // When the query string is a combination of dashes and dots, github API search doesn't work properly and doesn't find the image
+        // ex: xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime
+        // https://github.com/craftcms/cms/issues/3080
+        // To fix this, we need to replace the dashes with spaces in the search string and Github search can then find the image
+        // New string: xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime
+        // It doesn't effect other functions and works perfectly
+        if(query.contains("-")) {
+            query = query.replace('-', ' ');
```
afalko, line 90: I think you want to do `replaceAll`

afalko (changes requested):
Thanks for the change! Can you add a few unit tests for this and make sure we're replacing all of the dashes?

ksrisurapaneni:
@afalko : Which file to add the tests in?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
ksrisurapaneni: replaceAll fails with errors, but replace works just fine. :)

ksrisurapaneni: This is what we see when I use replaceAll:
```
Error:  COMPILATION ERROR : 
Error:  /home/runner/work/dockerfile-image-update/dockerfile-image-update/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java:[90,38] incompatible types: char cannot be converted to java.lang.String
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project dockerfile-image-update: Compilation failure
```

ksrisurapaneni: **Usage:** For our case, replace should work well since we are not using regex.

String replace(char oldChar, char newChar): It replaces all the occurrences of a oldChar character with newChar character. For e.g. "pog pance".replace('p', 'd') would return dog dance.

String replaceAll(String regex, String replacement): It replaces all the substrings that fits the given regular expression with the replacement String.

afalko: Ahh, excellent! I had a different model in my head

afalko:
> @afalko : Which file to add the tests in?

@ksrisurapaneni : The tests for this class are here: https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java

To make this change safer, I'm thinking we should have this enable if you pass a `--replace-dashes` command flag...or perhaps even `--replace-char-with "-: "`. That way we dont risk breaking existing flows.

ksrisurapaneni:
> > @afalko : Which file to add the tests in?
> 
> @ksrisurapaneni : The tests for this class are here: https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java
> 
> To make this change safer, I'm thinking we should have this enable if you pass a `--replace-dashes` command flag...or perhaps even `--replace-char-with "-: "`. That way we dont risk breaking existing flows.

@afalko : I don't know much Java to write new code. Could you perhaps write the test code for this? If not, I will have to look into this later.

justinharringa:
Hey there @ksrisurapaneni! I just merged https://github.com/salesforce/dockerfile-image-update/pull/212 and the bits are in [1.0.132](https://github.com/salesforce/dockerfile-image-update/releases/tag/dockerfile-image-update-1.0.132). Would you like to take a look to see if this solves the problem for you? If so, we could probably close this out.

We really appreciate your contributions and information here!

ksrisurapaneni:
Hi @justinharringa , thank you so much for fixing the issue. 1.0.132 seems to be working perfectly, closing this PR.

justinharringa:
Absolutely! Glad that worked out for you! Hope all is well!

PR #191, opened 2021-01-26
Hack GitHub search limitations

stephenc:
Hit a limit when using this against GCR repos where GitHub text search fails to find any `Dockerfile` with lines like:

```
FROM gcr.io/some-project/some-folder/some-image:tag
```

A github search for `FROM gcr.io/some-project` will find the repos, but a github search for `FROM gcr.io/some-project/some-folder` will not.

Hacky solution is to split on every second `/` and rely on the filtering effect of the initial `FROM`. This will give some false matches,  but the code that inspects the `Dockerfile`s for any potential changes will filter... only really a risk if you have 1000's of matches for that search

PR #212, opened 2021-03-27
Handle GitHub search limit: dashes in image name

### LINKED ISSUE #210: Image name with "-" in registry name not working
**Describe the bug**

Inconsistency in search methods across tool code is causing issues when using more complex image names and not letting it work as intended. For ex:

Looks like the tool finds the image and creates a PR successfully when the image is as follows:
```
FROM rv-python-runtime:1.0.0
```

But it doesn't find the image or create a PR  when the image is as follows:
```
FROM xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime:1.0.0
```

**To Reproduce**

Add this line to a test Dockerfile:

```
FROM xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime:1.0.0
```

Case 1: Run below command with image name as-is:

```
docker run --rm  -e git_api_token -e git_api_url salesforce/dockerfile-image-update:1.0.131 -o <org> -b main parent "xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime" 1.0.1 image_tag_store
```

Because the search is using Github API, it doesn't find the image name:

```
[main] INFO com.salesforce.dockerfileimageupdate.subcommands.impl.Parent - Finding Dockerfiles with the given image...
[main] INFO com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil - Number of files found for xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-runtime:0
```

Case 2: Replace dashes with spaces when running the below command:

```
docker run --rm  -e git_api_token -e git_api_url salesforce/dockerfile-image-update:1.0.131 -o <org> -b main parent "xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime" 1.0.1 image_tag_store
```

Since the tool initially uses Github API, the search works on replacing the "-"s with spaces. But looks like the forking process doesn't use github API and uses Java substring, so it goes on to fail then because it isn't able to find the image name with spaces in it and returns thhe below error:

```
[main] INFO com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil - Number of files found for xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime:2
[main] INFO com.salesforce.dockerfileimageupdate.process.GitHubPullRequestSender - Forking repositories...
[main] WARN com.salesforce.dockerfileimageupdate.process.GitHubPullRequestSender - Skipping <org>/foo because didn't find the image 'xyz asndnkdk ffwwwsw deeeereer.jfrog.io/rv python runtime' which required an update in path Dockerfile
```

**Expected behavior**

Ideally, we want such an image name search to work fine and create PRs. Currently, this is not possible with the tool.

**Additional context**

justinharringa:
Per #191 and #210, there are a couple of limitations in GitHub search
that effectively limit the ability of DFIU to find code results.

By combining known domain patterns and noting the limitations documented
in https://docs.github.com/en/github/searching-for-information-on-github/searching-code#considerations-for-code-search
this change does some treatment of image names. It treats the TLD and
remaining URL segments slightly different.

The domain will be split into parts by dashes while we'll attempt to
split the remaining url parts by slashes if they have dashes in them.
Meanwhile, we'll attempt to keep the search terms down to a minimum if
possible.

A fair number of working use cases based on manual testing can be
found in the unit tests but working and non-working tests are also
documented in #191 and #210. Some PRs have been created as a result of
this after these changes:
* https://github.com/justinharringa/somecontainers/pull/9
* https://github.com/justinharringa/somecontainers/pull/10
* https://github.com/justinharringa/somecontainers/pull/11
* https://github.com/justinharringa/somecontainers/pull/12
* https://github.com/justinharringa/somecontainers/pull/13

Fixes #210

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
@@ -116,7 +116,7 @@ public class GitHubImageSearchTermList {
      */
     static class ProcessingState {
         final List<String> terms = new ArrayList<>();
-        private StringBuffer termBuffer = new StringBuffer();
+        private final StringBuffer termBuffer = new StringBuffer();
 
         /**
          * Add to the current search term
@@ -139,7 +139,7 @@ public class GitHubImageSearchTermList {
          */
         void finalizeCurrentTerm() {
             terms.add(termBuffer.toString());
-            termBuffer = new StringBuffer();
+            termBuffer.setLength(0);
         }
     }
 }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
@@ -28,22 +28,47 @@ public class GitHubImageSearchTermList {
         ProcessingState state = processDomainPartOfImage(imageParts[0]);
         if (imageParts.length > 1) {
             for (int i = 1; i < imageParts.length - 1; i++) {
-                if (imageParts[i].contains("-")) {
-                    state.finalizeCurrentTerm();
-                } else {
-                    state.addToCurrentTerm("/");
-                }
-                state.addToCurrentTerm(imageParts[i]);
+                String imagePart = imageParts[i];
+                processIntermediateUrlSegment(state, imagePart);
             }
-            String leftoverTerm = state.getCurrentTerm();
-            if (leftoverTerm.contains("-")) {
-                state.finalizeCurrentTerm();
-            }
-            state.addToCurrentTerm("/");
-            state.addToCurrentTerm(imageParts[imageParts.length - 1]);
+            String finalImageSegment = imageParts[imageParts.length - 1];
+            processFinalUrlSegment(state, finalImageSegment);
         }
         state.finalizeCurrentTerm();
         return state.terms;
+    }
+
+    /**
+     * Intermediate URL segments will finalize the current search term if they have a dash and prep for
+     * a new search term. Otherwise, we'll add a slash and this {@code imagePart}.
+     *
+     * @param state processing state
+     * @param imagePart the current image segment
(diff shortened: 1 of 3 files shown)


justinharringa:
@ksrisurapaneni @stephenc I've taken a crack at resolving both of your use cases here. We REALLY appreciate you both reporting this and also taking a crack at it. I believe this puts us in a pretty good situation that also seems to handle current known use cases.

Fixes #211

ksrisurapaneni:
> @ksrisurapaneni @stephenc I've taken a crack at resolving both of your use cases here. We REALLY appreciate you both reporting this and also taking a crack at it. I believe this puts us in a pretty good situation that also seems to handle current known use cases.
> 
> Fixes #211

Thanks @justinharringa ! Waiting for this to be merged! :)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        state.addToCurrentTerm(domainParts[domainParts.length - 1]);
+    }
+
+    /**
+     * Helper class to store processing state of a search term or terms
+     */
+    static class ProcessingState {
+        final List<String> terms = new ArrayList<>();
+        private StringBuffer termBuffer = new StringBuffer();
```
afalko, line 119: Shouldn't this be `final`? It might be safer do use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/StringBuffer.html#delete(int,int) below?

justinharringa: Yeah, fair point - inbound change coming to `final` and `setLength(0)`

ksrisurapaneni: Looks good!

afalko (approved):

Issue #217, opened 2021-04-18
Unable to find image when there is a "." in the image name


ksrisurapaneni:
**Describe the bug**
Thank you for fixing #210. `salesforce/dockerfile-image-update:1.0.132` now works with "-" in the registry name. But looks like is unable to find when there is a "." in the image name. 

Looks like the tool finds the image and creates a PR successfully when the image is as follows:

```
FROM xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-3-runtime:1.0.0
```

But it doesn't find the image or create a PR when the image is as follows:

```
FROM xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-3.9-runtime:1.0.0
```

**To Reproduce**

Add this line to a test Dockerfile:

```
FROM xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-3.9-runtime:1.0.0
```

Run below command with image name as-is:
```
docker run --rm  -e git_api_token -e git_api_url salesforce/dockerfile-image-update:1.0.132 -o <org> -b main parent "xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-3.9-runtime" 1.0.1 image_tag_store
```

Error:

```
[main] INFO com.salesforce.dockerfileimageupdate.subcommands.impl.Parent - Finding Dockerfiles with the given image...
[main] INFO com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil - Number of files found for xyz-asndnkdk-ffwwwsw-deeeereer.jfrog.io/rv-python-3.9-runtime:0
```

**Expected behavior**
Ideally, we want such an image name search to work fine and create PRs. Currently, this is not possible with the tool.

The error again seems to be because the search is using Github API, it can't find the image name with "." in it. Can you please fix this? It will be really helpful, all our new images have this convention and will carry a ".".

Thanks,
Keerti

justinharringa:
Hi Keerti,

  I suspect this is an issue with the mixture of dashes and dots in the image name in GitHub Code Search. I'm aware of underscores and dots working well in GitHub Search (e.g. I suspect `rv_python_3.9_runtime` would work if images were named as such). While that doesn't really help your case unless you were to change your naming standard, I just wanted to document my thoughts/observations for readers here. 

  I'm not sure when I'll have time to poke at this. 

Hope all is well!

[...]

Issue #169, opened 2020-08-04
Handle cases where there are more than 1000 downstream consumers of an image


justinharringa:
GitHub Search [limits results to 1000 results per search](https://developer.github.com/v3/search/#about-the-search-api) but will give you the full total. We'll need a new strategy to deal with those instances where our [code search](https://developer.github.com/v3/search/#search-code) has too many results.

Some potential opportunities:
* split up searches to handle searches by org(s) when we know we have more than 1000 (create a set of targets)
  * potentially partition
* silly idea: if less than 2000 just sort ascending then descending which likely has further dragons and doesn't scale

Definitely should log this case so that it's obvious that we're missing some repos. This would be a hint that this should be run in a more scalable manner.

afalko:
There is no way to return paginated search?

npwolf:
Searching around, the way people have looked to get around this is by creating narrower search criteria, like slicing dates https://stackoverflow.com/q/37602893/13943777 . Doesn't appear to be any great solutions.

justinharringa:
Paginated search ends at 1000 (or 1020 if you leave the default). The date ranges don't work for code search (that example is for repository searches).

justinharringa:
Note that it would be possible to run this as a service if you can more efficiently store who the downstream consumers are and then just run the `child` command. That being said, it might be nice to somehow intelligently split out once we hit 1000 results but it seems that there are ultimately scaling limits in relying on GitHub search.

justinharringa:
Another potential fix to this would be for DFIU to try and do a best effort split of the orgs it finds if it found more than 1000.

Essentially:
```
orgsToExclude = []
while searchResults.length() > 1000
    orgWithMostHits = getOrgWithMostHits(searchResults)
    runDFIUForOrg(orgWithMostHits)
    orgsToExclude.add(orgWithMostHits)
    searchResults = getNewSearchResults(orgsToExclude)
```

PR #240, opened 2021-06-15
fix for github search on image name

rv-msaraf:


ksrisurapaneni:
Related to this issue: https://github.com/salesforce/dockerfile-image-update/issues/217

.github/workflows/cd.yml
```diff
@@ -27,104 +27,104 @@ jobs:
...
-          encrypted_96e73e3cb232_key: ${{ secrets.encrypted_96e73e3cb232_key }}
-          GPG_KEY_NAME: ${{ secrets.GPG_KEY_NAME }}
-          GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
-          CI_DEPLOY_USER: ${{ secrets.CI_DEPLOY_USER }}
-          CI_DEPLOY_PASSWORD: ${{ secrets.CI_DEPLOY_PASSWORD }}
-          NEW_PATCH_VERSION: ${{ needs.get-next-version.outputs.version }}
-        run: make deploy
+      # - name: Upload itest module
```
justinharringa, line 30: Could you help me understand why you're doing this? Ideally, you shouldn't need to modify our CI.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -81,6 +81,12 @@ public GHRepository getRepo(String repoName) throws IOException {
         if (image.substring(image.lastIndexOf(' ') + 1).length() <= 1) {
             throw new IOException("Invalid image name.");
         }
+
+        // replace dots with spaces in image name, for github search to work
+        if(image.contains(".")) {
```
justinharringa, line 86: This is a very broad change that has some pretty bad consequences. The worst consequence is that this can expand the search space which will trigger #169 

You might want to take a look at how we handled this for other portions of the images in #212. It'd also be possible to gate this functionality behind a new command-line flag but I still think it'd be best to incorporate a similar strategy to #212

justinharringa (changes requested):
Hey there! Thank you for the contribution! I have some questions / suggestions when you get a chance.

.github/workflows/cd.yml
ksrisurapaneni: Hi, this was just for our local, please ignore changes to CI/CD.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
ksrisurapaneni: Hi @justinharringa , we were wondering if you could create code similar to #212 for this. The changes in this PR are just to show how we are resolving the issue on our end for now by creating a new image locally. Thanks!

justinharringa: No worries. 😄  I would like to get some time to do this but work has been quite busy lately and finding time in my free time has also been a bit of a challenge. If you'd like to take a shot at it, that'd be awesome. But also appreciate you helping illustrate what has worked well for you. 😄

PR #245, opened 2021-06-16
simple fix for issue 217

ksrisurapaneni:

PR #251, opened 2021-06-19
Split image name if there are dots

### LINKED ISSUE #217: Unable to find image when there is a "." in the image name

justinharringa:
As in #212, mixing dots and dashes makes GitHub Code Search sad.
Since it's less likely to encounter dots than dashes in an image name,
and because we've covered previous segments, we'll split the image
name by dots.

Fixes #217 #245

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/search/GitHubImageSearchTermList.java
@@ -7,7 +7,7 @@ import java.util.List;
 
 public class GitHubImageSearchTermList {
     /**
-     * We need to split out search terms to accomodate how GitHub code search works
+     * We need to split out search terms to accommodate how GitHub code search works
      * https://docs.github.com/en/github/searching-for-information-on-github/searching-code#considerations-for-code-search
      *
      * Essentially we'll split out any dashes in the registry domain and split out any url segments with dashes by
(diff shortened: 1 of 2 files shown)


justinharringa:
Did some additional verification beyond the tests. I was able to reproduce @ksrisurapaneni's #217 in the latest version and was able to get DFIU to create a PR with this. The result is: https://github.com/justinharringa/somecontainers/pull/18

ksrisurapaneni:
Waiting for the merge to happen! :)

afalko (approved):
Looks great!

PR #276, opened 2021-10-13
Workaround for the GitHub Search API limit

avimanyum:
Currently, due to a GIT API search limit of 1000, DFIU can only send PRs to up to 1000 downstream repos when a new version for a base image is available. With this change, if the number of downstream repos consuming a base image is more than 1000, the Git org with the most number of consumers is excluded from the search space. DFIU is run separately on that individual org while another search is executed while excluding that org. This process is repeated, by adding orgs recursively with the most number of consumers to the exclude list to reduce the search space till the resulting number of consumers is less than 1000. For each of the orgs that are added to the exclusion list, DFIU is run individually on each of those orgs.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -48,12 +48,18 @@ public class All implements ExecutableWithNamespace {
         Multimap<String, String> pathToDockerfilesInParentRepo = ArrayListMultimap.create();
 
         Set<Map.Entry<String, JsonElement>> imageToTagStore = parseStoreToImagesMap(ns.get(Constants.STORE));
+        Integer gitApiSearchLimit;
+        if (ns.get(Constants.GIT_API_SEARCH_LIMIT) == null || Integer.parseInt(ns.get(Constants.GIT_API_SEARCH_LIMIT)) > Constants.GIT_API_SEARCH_LIMIT_NUMBER) {
+            gitApiSearchLimit = Constants.GIT_API_SEARCH_LIMIT_NUMBER;
+        } else {
+            gitApiSearchLimit = Integer.parseInt(ns.get(Constants.GIT_API_SEARCH_LIMIT));
+        }
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
             String image = imageToTag.getKey();
             log.info("Repositories with image {} being forked.", image);
             imageToTagMap.put(image, imageToTag.getValue().getAsString());
             List<PagedSearchIterable<GHContent>> contentsWithImage =
-                    this.dockerfileGitHubUtil.findFilesWithImage(image, ns.get(Constants.GIT_ORG));
+                    this.dockerfileGitHubUtil.findFilesWithImage(image, ns.get(Constants.GIT_ORG), null, gitApiSearchLimit);
             for (int i = 0; i < contentsWithImage.size(); i++) {
                 forkRepositoriesFound(pathToDockerfilesInParentRepo,
                         imagesFoundInParentRepo, contentsWithImage.get(i), image);
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -51,7 +51,9 @@ public class All implements ExecutableWithNamespace {
         Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
         List<IOException> exceptions = new ArrayList<>();
         Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
-        orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
+        if (!ns.get(Constants.GIT_ORG).toString().isEmpty()) {
+            orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
+        }
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
             String image = imageToTag.getKey();
             log.info("Repositories with image {} being forked.", image);
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -51,9 +51,7 @@ public class All implements ExecutableWithNamespace {
         Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
         List<IOException> exceptions = new ArrayList<>();
         Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
-        if (!ns.get(Constants.GIT_ORG).toString().isEmpty()) {
-            orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
-        }
+        orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
             String image = imageToTag.getKey();
             log.info("Repositories with image {} being forked.", image);
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -48,22 +48,25 @@ public class All implements ExecutableWithNamespace {
         Multimap<String, String> pathToDockerfilesInParentRepo = ArrayListMultimap.create();
 
         Set<Map.Entry<String, JsonElement>> imageToTagStore = parseStoreToImagesMap(ns.get(Constants.STORE));
-        Integer gitApiSearchLimit;
-        if (ns.get(Constants.GIT_API_SEARCH_LIMIT) == null || Integer.parseInt(ns.get(Constants.GIT_API_SEARCH_LIMIT)) > Constants.GIT_API_SEARCH_LIMIT_NUMBER) {
-            gitApiSearchLimit = Constants.GIT_API_SEARCH_LIMIT_NUMBER;
-        } else {
-            gitApiSearchLimit = Integer.parseInt(ns.get(Constants.GIT_API_SEARCH_LIMIT));
-        }
+        Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
+        List<IOException> exceptions = new ArrayList<>();
+        Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
+        orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
             String image = imageToTag.getKey();
             log.info("Repositories with image {} being forked.", image);
             imageToTagMap.put(image, imageToTag.getValue().getAsString());
-            List<PagedSearchIterable<GHContent>> contentsWithImage =
-                    this.dockerfileGitHubUtil.findFilesWithImage(image, ns.get(Constants.GIT_ORG), null, gitApiSearchLimit);
-            for (int i = 0; i < contentsWithImage.size(); i++) {
-                forkRepositoriesFound(pathToDockerfilesInParentRepo,
-                        imagesFoundInParentRepo, contentsWithImage.get(i), image);
-            }
+            Optional<List<PagedSearchIterable<GHContent>>> contentsWithImage =
+                    this.dockerfileGitHubUtil.findFilesWithImage(image, orgsToIncludeInSearch, gitApiSearchLimit);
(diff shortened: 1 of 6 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -83,14 +83,27 @@ public GHRepository getRepo(String repoName) throws IOException {
...
+            Integer gitApiSearchLimit) throws IOException {
         GHContentSearchBuilder search = gitHubUtil.startSearch();
         // Filename search appears to yield better / more results than language:Dockerfile
         // Root cause: linguist doesn't currently deal with prefixes of files:
         // https://github.com/github/linguist/issues/4566
         search.filename("Dockerfile");
-        if (org != null) {
-            search.user(org);
+        if (!orgsToIncludeOrExclude.isEmpty()) {
+            StringBuilder includeOrexcludeOrgsQuery = new StringBuilder();
```
justinharringa, line 96:
```suggestion
            StringBuilder includeOrExcludeOrgsQuery = new StringBuilder();
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -48,14 +48,27 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
         Multimap<String, String> pathToDockerfilesInParentRepo = ArrayListMultimap.create();
 
         Set<Map.Entry<String, JsonElement>> imageToTagStore = parseStoreToImagesMap(ns.get(Constants.STORE));
+        Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
+        List<IOException> exceptions = new ArrayList<>();
+        Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
```
justinharringa, line 53: Are we wanting a `Set` instead? Should it be unmodifyable? What is the `value` here? Should this be a `Set` with typed `IncludedOrgs` / `ExcludedOrgs` or something like that?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -48,14 +48,27 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+        orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
```
justinharringa, line 54: What if `GIT_ORG` wasn't specified? Does this work properly?

justinharringa:
@avimanyum Thanks! This is awesome! Could you edit the description of what the net change of this PR does? It seems to appear that it will potentially exclude one org with the top hits but I haven't given it a super deep look.

justinharringa (changes requested):
@avimanyum Thanks! This is awesome! Could you edit the description of what the net change of this PR does? It seems to appear that it will potentially exclude one org with the top hits but I haven't given it a super deep look.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: We wanted to have a map where the keys would be Org names and the value would be a Boolean value. True indicates the the org should be included in the search, and False indicates that it should not be included in the search. Using a map also ensures that the same org cannot be included and excluded at the same time.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: > @avimanyum Thanks! This is awesome! Could you edit the description of what the net change of this PR does? It seems to appear that it will potentially exclude one org with the top hits but I haven't given it a super deep look.

Added the description [here](https://github.com/salesforce/dockerfile-image-update/pull/276#issue-757738367). Please let me know if that looks accurate. Is there some other place you wanted me to add this to? A

avimanyum: Thanks for catching this. Will add a check here

afalko (dismissed):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
justinharringa: Gotcha. I guess I could read this a few ways. Since the variable is named `orgsToIncludeInSearch` I don't know that I'd immediately think of the include/exclude so that's where I was wondering if it deems a formal class for coherence. I guess the other way would be to add a comment for clarity.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
justinharringa: Were you still going to add a check?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -83,14 +84,29 @@ public GHRepository getRepo(String repoName) throws IOException {
...
         search.filename("Dockerfile");
-        if (org != null) {
-            search.user(org);
+        if (!orgsToIncludeOrExclude.isEmpty()) {
+            StringBuilder includeOrExcludeOrgsQuery = new StringBuilder();
+            for (Map.Entry<String, Boolean> org : orgsToIncludeOrExclude.entrySet()){
+                if (org.getKey() != null) {
+                    if (org.getValue()) {
+                        log.info("Including the org {} in the search.", org.getKey());
+                        includeOrExcludeOrgsQuery.append(String.format("user:%s ", org.getKey()));
```
justinharringa, line 102: why not simply do `search.user(org.getKey())`?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -83,14 +84,29 @@ public GHRepository getRepo(String repoName) throws IOException {
...
+                    } else {
+                        log.info("Excluding the org {} from the search.", org.getKey());
+                        includeOrExcludeOrgsQuery.append(String.format("-org:%s ", org.getKey()));
+                    }
+                }
+            }
+            search.q(includeOrExcludeOrgsQuery.toString());
```
justinharringa, line 109: Do we need to check to make sure we don't exceed any search term limits?

justinharringa (commented):
Looks like you need to rebase. Added some more comments. Looks like there are some unresolved comments from Andrey and I.

Thanks for the description add! I think that makes sense.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
avimanyum: There can be multiple orgs that will need to be appended to the search query. From my understanding `search.user(org)` internally reslolves to `search.q(user:org)`. If we call call `search.user(org)` wouldn't the search query keep getting overwritten with every call and have only the last org added? Instead, using the `search.q` method allows us to construct the query string with all orgs.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
avimanyum: Can you please point me to some documentation about search term limits?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: When the method call happens here, the map will only have an org to include in the search. Hence, I chose to name it `orgsToIncludeInSearch`. In the actual method definition, this gets stored in a variable `Map<String, Boolean> orgsToIncludeOrExclude` since this data structure can contain orgs to include or exclude depending on where it is being invoked from.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: Yes, added a check.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
justinharringa: https://docs.github.com/en/rest/reference/search#limitations-on-query-length

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
justinharringa: If you dig into `search.q()` you'll notice that it's just appending search terms to a `GHSearchBuilder`. If it were the case that it was overriding, you'd have a similar problem with `search.q()` here overriding `search.filename()` above.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
avimanyum: Thanks @justinharringa!
Do we need to add an additional check here or can we rely on the check that Github will do internally and return a "Validation failed" error if the limit is exceeded?

justinharringa: I have not spotted anything in the github-api Java library that does this check but I'm not sure. I suspect that you'd likely get a failure from the actual GitHub API in this case.

Seems like it would be worth testing. A simple way to test would be to simply try and build a `main()` that adds 500 search terms of length beyond the limit and see what happens.

avimanyum: I added 500 search terms to make sure that it exceeds the limit mentioned [here](https://docs.github.com/en/rest/reference/search#limitations-on-query-length) and got the following error 
`Caused by: org.kohsuke.github.HttpException: {"message":"Validation Failed","errors":[{"message":"The search is longer than 128 characters.","resource":"Search","field":"q","code":"invalid"}],"documentation_url":"https://docs.github.com/enterprise/3.1/v3/search/"}`

This validation is handled at the github api level, so maybe we do not need to add any more validations here.

justinharringa: Ouch. That error shows that it only allows 128 characters. Is the thought that there would be some follow-on work to split out the search to individual org searches? It seems like it may not be hard to hit this limit for the overall search if we end up with something like this:
`filename:Dockerfile -org:somelonguser -org:someotherreallylonguser -org:anotherreallylonguser -org:ohnothisoneisfartoolongandwefailed`

avimanyum: Running the search for every individual org will increase # of gh api calls a lot. But reaching a 128 chars limit would not be very hard like you mentioned.
CC @afalko

afalko: I don't think we need to do a search for each individual org unless each of them are 90 long. I think we'll need do the following if the total length of orgs + the Dockerfile search is >120:
1) Sort orgs by length
2) Build list of smallest length orgs until you have as many as will fit under the 120 limit
3) Call search with those
4) Repeat step #2 until you don't have any orgs remaining. 
We'll need a function that calls search again with another set of orgs if

justinharringa: I'll let you 2 make the determination on follow-up or not - would be good to check one of our known large orgs to get an idea of how impactful this is. Note that `github-api` does try and respect rate limiting in some cases with retries (although it's simple sleep where this does occur). I believe that this is one of those cases so that certainly helps with GH API calls.

afalko (approved):
Despite the search space size limitation, this buys some time for a better solution.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
afalko: @justinharringa : we determined to proceed forward without the above and create a follow up change to either do the above or take another approach all together.

justinharringa (approved):

PR #284, opened 2021-12-17
Add new flag to parent

avimanyum:
This PR adds a new flag to the parent command that allows us to skip the PR creation operations. If this flag is set to "true", DFIU will only update the image tag store and exit without creating PRs to any of the downstream repos. If the flag is set to "false", DFIU continues behaving the same way it does now. By default, the flag is set to false.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
@@ -78,6 +78,10 @@ public class CommandLine {
                 .help("regex of repository names to exclude from pull request generation");
         parser.addArgument("-B")
                 .help("additional body text to include in pull requests");
+        parser.addArgument("-s")
+                .type(Boolean.class)
+                .setDefault(false)
+                .help("Only update image tag store. Skip creating PRs");
         return parser;
     }
 
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
@@ -78,7 +78,7 @@ public class CommandLine {
                 .help("regex of repository names to exclude from pull request generation");
         parser.addArgument("-B")
                 .help("additional body text to include in pull requests");
-        parser.addArgument("-s")
+        parser.addArgument("-s", "--" + Constants.SKIP_PR_CREATION)
                 .type(Boolean.class)
                 .setDefault(false)
                 .help("Only update image tag store. Skip creating PRs");
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
@@ -52,7 +52,7 @@ public class Parent implements ExecutableWithNamespace {
 
         if (ns.get(Constants.SKIP_PR_CREATION)) {
             log.info(String.format("Since the flag %s is set to True, the PR creation steps will " +
-                    "be skipped. ", ns.get(Constants.SKIP_PR_CREATION)));
+                    "be skipped. ", Constants.SKIP_PR_CREATION));
             System.exit(0);
         }
         GitHubPullRequestSender pullRequestSender =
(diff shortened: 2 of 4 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -78,6 +78,10 @@ static ArgumentParser getArgumentParser() {
                 .help("regex of repository names to exclude from pull request generation");
         parser.addArgument("-B")
                 .help("additional body text to include in pull requests");
+        parser.addArgument("-s", "--" + Constants.SKIP_PR_CREATION)
```
justinharringa, line 81: Would `--dry-run` make sense?

justinharringa: Actually, this seems to be of a different intent than just skipping PR creation as well. Should this be a different command altogether?

afalko: What would you suggest the subcommand name be @justinharringa ? 

dry-run isn't good because we intend to change our datastore, which is a mutation. 

Naming is always hard, but we can also change this flag to be `--only-persist-tag-store`.

I'm honestly partial to keeping what @avimanyum has right now.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -50,6 +50,11 @@ public void execute(final Namespace ns, DockerfileGitHubUtil dockerfileGitHubUti
         log.info("Updating store...");
         this.dockerfileGitHubUtil.getGitHubJsonStore(ns.get(Constants.STORE)).updateStore(img, tag);
 
+        if (ns.get(Constants.SKIP_PR_CREATION)) {
+            log.info(String.format("Since the flag %s is set to True, the PR creation steps will " +
```
afalko, line 54: Not need to `String.format`

Do this instead: 
`log.info("Since the flag {} is set to True, the PR creation steps will be skipped.", Constants.SKIP_PR_CREATION)`

Also, notice you have a trailing space after the period as well.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -50,6 +50,11 @@ public void execute(final Namespace ns, DockerfileGitHubUtil dockerfileGitHubUti
         log.info("Updating store...");
         this.dockerfileGitHubUtil.getGitHubJsonStore(ns.get(Constants.STORE)).updateStore(img, tag);
 
+        if (ns.get(Constants.SKIP_PR_CREATION)) {
```
afalko, line 53: Can you add a utest? We need to ensure that if somebody adds other processing that is non-PR related isn't surprised.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
justinharringa: Perhaps a mode called `update-store`? Seems that would also be a simple change.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -50,6 +50,11 @@ public void execute(final Namespace ns, DockerfileGitHubUtil dockerfileGitHubUti
...
+            log.info(String.format("Since the flag %s is set to True, the PR creation steps will " +
+                    "be skipped. ", Constants.SKIP_PR_CREATION));
+            System.exit(0);
```
justinharringa, line 56: Would a `return` be more natural here?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
avimanyum: To keep this consistent between `child` and `parent` subcommands, we can add a similar check in the `child` sub command execution as well.

afalko (approved):
Thanks!

PR #316, opened 2022-02-21
Catch exceptions while finding files with image

avimanyum:
For the `all` subcommand, if there is an exception thrown while searching for one image, that exception should be handled so that the `all` subcommand can continue on to the next image on the image tag store.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -45,6 +45,7 @@ public class All implements ExecutableWithNamespace {
             // the org gets included in the search query.
             orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
         }
+        List<String> imagesThatCouldNotBeProcessed = new ArrayList<>();
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
             String image = imageToTag.getKey();
             String tag = imageToTag.getValue().getAsString();
@@ -66,8 +67,11 @@ public class All implements ExecutableWithNamespace {
                         log.error("Could not send pull request.");
                     }
                 });
+            } else {
+                imagesThatCouldNotBeProcessed.add(image);
             }
         }
+        log.info("The list of images for which a PR could not be created are: {}", imagesThatCouldNotBeProcessed);
     }
 
     protected void loadDockerfileGithubUtil(DockerfileGitHubUtil _dockerfileGitHubUtil) {
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -71,7 +71,7 @@ public class All implements ExecutableWithNamespace {
                 imagesThatCouldNotBeProcessed.add(image);
             }
         }
-        log.info("The list of images for which a PR could not be created are: {}", imagesThatCouldNotBeProcessed);
+        log.warn("The list of images for which a PR could not be created are: {}", imagesThatCouldNotBeProcessed);
     }
 
     protected void loadDockerfileGithubUtil(DockerfileGitHubUtil _dockerfileGitHubUtil) {
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -15,6 +15,7 @@ import com.salesforce.dockerfileimageupdate.process.*;
 import com.salesforce.dockerfileimageupdate.subcommands.ExecutableWithNamespace;
 import com.salesforce.dockerfileimageupdate.utils.Constants;
 import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
+import com.salesforce.dockerfileimageupdate.utils.ProcessingErrors;
 import com.salesforce.dockerfileimageupdate.utils.PullRequests;
 import net.sourceforge.argparse4j.inf.Namespace;
 import org.kohsuke.github.*;
@@ -23,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
 
 @SubCommand(help="updates all repositories' Dockerfiles",
         requiredParams = {Constants.STORE})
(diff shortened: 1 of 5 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -66,8 +67,11 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                         log.error("Could not send pull request.");
                     }
                 });
+            } else {
+                imagesThatCouldNotBeProcessed.add(image);
             }
         }
+        log.info("The list of images for which a PR could not be created are: {}", imagesThatCouldNotBeProcessed);
```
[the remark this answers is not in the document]
afalko: `log.warn`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -110,45 +110,50 @@ public GHRepository getRepo(String repoName) throws IOException {
...
+                log.info("The number of files returned is greater than the git API search limit"
+                        + " of {}. The orgs with the maximum number of hits will be recursively removed"
+                        + " to reduce the search space. For every org that is excluded, a separate "
+                        + "search will be performed specific to that org.", gitApiSearchLimit);
+                return getSearchResultsExcludingOrgWithMostHits(image, files, orgsToIncludeOrExclude, gitApiSearchLimit);
+            }
+            List<PagedSearchIterable<GHContent>> filesList = new ArrayList<>();
+            filesList.add(files);
+            return Optional.of(filesList);
+        } catch (GHException | HttpException e) {
```
afalko, line 153: I think it is best to keep this method as is and catch these exceptions in the `All` class. The reason is that now there is side-effect behavior on our Parent sub-command, which should be expected to fail if we find an image that cannot be looked up.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -66,8 +67,11 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                         log.error("Could not send pull request.");
                     }
                 });
+            } else {
+                imagesThatCouldNotBeProcessed.add(image);
```
afalko, line 71: a) I like that we are now recording the images that didn't return results to this error summary list. 
b) We should move move the try catch from the util class to per my comment there. If you do that, I recommend having the try block start at line 52 and go until line 59.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -66,8 +67,11 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                         log.error("Could not send pull request.");
```
afalko, line 67: We should append to imagesThatCouldNotBeProcessed here as well

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -45,6 +45,7 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
             // the org gets included in the search query.
             orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
         }
+        List<String> imagesThatCouldNotBeProcessed = new ArrayList<>();
```
afalko, line 48: Can you also add an `AtomicInteger` counter of images successfully processed, so we can print a summary of number of images processed vs. failed. e.g. `11 of 18 (61%) parent images processed. 7 images failed to process, see error summary below.`

afalko: For the error summary, let's create a POJO...
`List<ProcessingErrors> imagesThatCouldNotBeProcessed = new LinkedList<>();`
where `ProcesseingErrors` is: 
`ProcessingErrors(String imageNameAndTag, Optional<Exception> failure)`

afalko (changes requested):
We need to move the try catch clause to the All class

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -45,29 +47,62 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
             // the org gets included in the search query.
             orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
         }
+        List<ProcessingErrors> imagesThatCouldNotBeProcessed = new LinkedList<>();
+        AtomicInteger numberOfImagesToProcess = new AtomicInteger();
+        AtomicInteger numberOfImagesFailedToProcess = new AtomicInteger();
```
afalko, line 52: Not necessary, you have `imagesThatCouldNotBeProcessed.size()` ;)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -45,29 +47,62 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+                log.error("Could not perform Github search for the image {}. Trying to proceed...", image);
+                processErrors(image, tag, e, imagesThatCouldNotBeProcessed, numberOfImagesFailedToProcess);
             }
         }
+        printSummary(imagesThatCouldNotBeProcessed, numberOfImagesToProcess, numberOfImagesFailedToProcess);
+    }
+
+    protected void printSummary(List<ProcessingErrors> imagesThatCouldNotBeProcessed, AtomicInteger numberOfImagesToProcess, AtomicInteger numberOfImagesFailedToProcess) {
+        AtomicInteger numberOfImagesSuccessfullyProcessed = new AtomicInteger(numberOfImagesToProcess.get() - numberOfImagesFailedToProcess.get());
+        log.warn("The total number of images to process from image tag store: {}", numberOfImagesToProcess.get());
```
afalko, line 88:
```suggestion
        log.info("The total number of images to process from image tag store: {}", numberOfImagesToProcess.get());
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -45,29 +47,62 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+        log.warn("The total number of images that were successfully processed: {}", numberOfImagesSuccessfullyProcessed.get());
```
afalko, line 89:
```suggestion
        log.info("The total number of images that were successfully processed: {}", numberOfImagesSuccessfullyProcessed.get());
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -45,29 +47,62 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+        log.warn("The total number of images that failed to be processed: {}. The following list shows the images that could not be processed.", numberOfImagesFailedToProcess.get());
```
afalko, line 90:
```suggestion
        if (numberOfImagesFailedToProcess.isPresent() && numberOfImagesFailedToProcess.get().size() > 0) {
        log.warn("The total number of images that failed to be processed: {}. The following list shows the images that could not be processed.", numberOfImagesFailedToProcess.get());
        }
```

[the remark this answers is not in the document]
afalko: Oh actually, you can just print the size if it is present since I asked to remove this var

afalko (approved):

PR #320, opened 2022-02-23
Catching IOException when the image name is invalid

avimanyum:
On running the All subcommand, if there is an image in the tag store that has an invalid image name, the findFilesWithImage method throws an IOException [here](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L86). However, in the All subcommand execute path we are not catching this [exception](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java#L77). As a result, if the image tag store has one entry with an invalid image name, the all subcommand will exit once it hits that entry and not process the following entries. 
This fix allows us to catch these exceptions and move on to processing the following images in the image tag store.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -74,7 +74,7 @@ public class All implements ExecutableWithNamespace {
                         }
                     });
                 }
-            } catch (GHException | HttpException e){
+            } catch (GHException | IOException e){
                 log.error("Could not perform Github search for the image {}. Trying to proceed...", image);
                 processErrors(image, tag, e, imagesThatCouldNotBeProcessed);
             }


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -74,7 +74,7 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                         }
                     });
                 }
-            } catch (GHException | HttpException e){
```
afalko, line 77: You are sure we don't still also need `HttpException`? I guess `IOException` is a superset :thinking:

afalko (approved):

afalko (changes requested):
I think you forgot to clean up the import on `httpexception`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: Yes, it is!

justinharringa:
Could you provide more information on what the scenario is here? It's unclear what is happening and why this should be an exception rather than a validation.

avimanyum:
> Could you provide more information on what the scenario is here? It's unclear what is happening and why this should be an exception rather than a validation.

Added [here](https://github.com/salesforce/dockerfile-image-update/pull/320#issue-1148690250)

justinharringa:
> > Could you provide more information on what the scenario is here? It's unclear what is happening and why this should be an exception rather than a validation.
> 
> Added [here](https://github.com/salesforce/dockerfile-image-update/pull/320#issue-1148690250)

What do we mean by an "invalid image name"? It doesn't exist? The format is not correct? Examples would be useful if you could redact them.

Thanks again!

afalko (approved):
We are good. We `*` add exception

avimanyum:
> > > Could you provide more information on what the scenario is here? It's unclear what is happening and why this should be an exception rather than a validation.
> > 
> > 
> > Added [here](https://github.com/salesforce/dockerfile-image-update/pull/320#issue-1148690250)
> 
> What do we mean by an "invalid image name"? It doesn't exist? The format is not correct? Examples would be useful if you could redact them.
> 
> Thanks again!

Yeah sure. This is the [logic](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L110-L112) we have to determine the validity of an image name. An example of an invalid image name would be `a:a`. This happened to be in a tag store that the canary instance was using. That's how I encountered this bug :)

justinharringa:
weird... cool thanks for the info 😄 might be helpful for others too

PR #321, opened 2022-02-25
Handling error messaging when PR creation fails with Invalid code error

avimanyum:
There are cases when the PR creation fails and the exception message thrown looks like this:
`{"message":"Validation Failed","errors":[{"resource":"PullRequest","field":"head","code":"invalid"}],"documentation_url":"https://docs.github.com/enterprise/3.1/rest/reference/pulls#create-a-pull-request"}`

In such case, [this](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java#L111) line will throw a NullPointerException since the JsonArray does not have a field called `message`. Unfortunately, since this error is thrown inside the `catch` block, it is not handled and essentially terminates the process. 

This PR checks for the existence of this key inside the Json object before trying to get it. If it does not find it there, it assumes that the error was due to the head having invalid code (error message shown above). In such a scenario, the [method](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java#L96) returns an error code of 1, so that we can proceed to safely deleting the forked repo [here](https://github.com/salesforce/dockerfile-image-update/blob/master/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L340)

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -108,12 +108,17 @@ public class GitHubUtil {
             log.warn("Handling error with pull request creation... {}", e.getMessage());
             JsonElement root = JsonParser.parseString(e.getMessage());
             JsonArray errorJson = root.getAsJsonObject().get("errors").getAsJsonArray();
-            String error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            String error = "";
+            if (errorJson.get(0).getAsJsonObject().has("message")) {
+                error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            } else {
+                error = "PR head has invalid code.";
+            }
             log.info("error: {}", error);
             if (error.startsWith("A pull request already exists")) {
                 log.info("NOTE: {} New commits may have been added to the pull request.", error);
                 return 0;
-            } else if (error.startsWith("No commits between")) {
+            } else if (error.startsWith("No commits between") || error.startsWith("PR head has invalid code.")) {
                 log.warn("NOTE: {} Pull request was not created.", error);
                 return 1;
             } else {
--- a/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
+++ b/dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
@@ -120,6 +120,19 @@ public class GitHubUtilTest {
         assertEquals(gitHubUtil.createPullReq(origRepo, "branch", forkRepo, "title", "body"), 1);
         verify(origRepo, times(1)).createPullRequest(eq("title"), eq("owner:branch"), eq("master"), eq("body"));
     }
+    @Test
+    public void testCreatePullReq_errorCase1_withInvalidCode() throws Exception {
+        GitHub github = mock(GitHub.class);
+        GitHubUtil gitHubUtil = new GitHubUtil(github);
+        GHRepository origRepo = mock(GHRepository.class);
+        when(origRepo.getDefaultBranch()).thenReturn("master");
+        when(origRepo.createPullRequest(eq("title"), eq("owner:branch"), eq("master"), eq("body")))
+                .thenThrow(new IOException("{\"message\":\"Validation Failed\",\"errors\":[{\"resource\":\"PullRequest\",\"field\":\"head\",\"code\":\"invalid\"}],\"documentation_url\":\"https://developer.github.com/enterprise/2.6/v3/pulls/#create-a-pull-request\"}"));
+        GHRepository forkRepo = mock(GHRepository.class);
+        when(forkRepo.getOwnerName()).thenReturn("owner");
+        assertEquals(gitHubUtil.createPullReq(origRepo, "branch", forkRepo, "title", "body"), 1);
+        verify(origRepo, times(1)).createPullRequest(eq("title"), eq("owner:branch"), eq("master"), eq("body"));
+    }
 
     @Test
     public void testTryRetrievingRepository() throws Exception {


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -108,12 +108,17 @@ public int createPullReq(GHRepository origRepo, String branch,
             log.warn("Handling error with pull request creation... {}", e.getMessage());
             JsonElement root = JsonParser.parseString(e.getMessage());
             JsonArray errorJson = root.getAsJsonObject().get("errors").getAsJsonArray();
-            String error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            String error = "";
+            if (errorJson.get(0).getAsJsonObject().has("message")) {
+                error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            } else {
+                error = "PR head has invalid code.";
```
afalko, line 115: What does "invalid code" mean? What action should be taken to resolve if any?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -108,12 +108,17 @@ public int createPullReq(GHRepository origRepo, String branch,
...
+            }
             log.info("error: {}", error);
             if (error.startsWith("A pull request already exists")) {
                 log.info("NOTE: {} New commits may have been added to the pull request.", error);
                 return 0;
-            } else if (error.startsWith("No commits between")) {
+            } else if (error.startsWith("No commits between") || error.startsWith("PR head has invalid code.")) {
```
afalko, line 121: Repeated error message. Let's make it a constant in this class

afalko (changes requested):
LGTM, but I think we should improve the error message.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
avimanyum: The error message is trying to communicate that the field/keyword called "head" contains invalid data. When i saw this message, I manually deleted the fork and the next run was successful. This change will also do the same - deleted the fork.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -29,6 +29,7 @@
 public class GitHubUtil {
     private static final Logger log = LoggerFactory.getLogger(GitHubUtil.class);
     public static final String NO_BRANCH_WARN_FORMAT = "Couldn't find branch `%s` in repo `%s`. Waiting a second...";
+    public static final String PR_INVALID_CODE = "The PR head has invalid data.";
```
afalko, line 32: I still don't think a user or admin will understand this error message. Can we hint at what needs to be done to fix the PR?

afalko (approved):

justinharringa:
Hmm... Are there other scenarios that end up taking this path? My concern is that when you delete a fork, the API will lag by hours and may select a fork that is no longer there. Additionally, I believe this may have an effect on open PRs.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -108,12 +110,18 @@ public int createPullReq(GHRepository origRepo, String branch,
...
+            } else {
+                // This case usually happens when the PR head has invalid data. Deleting the forked repo resolves it.
+                error = PR_INVALID_CODE;
+            }
             log.info("error: {}", error);
             if (error.startsWith("A pull request already exists")) {
                 log.info("NOTE: {} New commits may have been added to the pull request.", error);
                 return 0;
-            } else if (error.startsWith("No commits between")) {
+            } else if (error.startsWith("No commits between") || error.startsWith(PR_INVALID_CODE)) {
```
cjneasbi, line 124: Create a list of error prefixes and use stream operations

```
List<String> prefixes = Arrays.asList(PR_INVALID_CODE, "No commits between")

if(prefixes.anyMatch(prefix -> errors.startsWith(prefix))
```
This way if you add more prefixes the if statement doesn't change.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -108,12 +110,18 @@ public int createPullReq(GHRepository origRepo, String branch,
             log.warn("Handling error with pull request creation... {}", e.getMessage());
             JsonElement root = JsonParser.parseString(e.getMessage());
             JsonArray errorJson = root.getAsJsonObject().get("errors").getAsJsonArray();
-            String error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            String error = "";
+            if (errorJson.get(0).getAsJsonObject().has("message")) {
```
cjneasbi, line 114: Store the output of `errorJson.get(0).getAsJsonObject()` as a local variable.  Call `has` and `get` on that variable.  This removes the need to call `get(0).getAsJsonObject()` twice.

cjneasbi: Why are we ignoring other errors if multiple were returned?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -108,12 +110,18 @@ public int createPullReq(GHRepository origRepo, String branch,
...
+                error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
+            } else {
+                // This case usually happens when the PR head has invalid data. Deleting the forked repo resolves it.
+                error = PR_INVALID_CODE;
```
cjneasbi, line 118: You don't need this else.  Just assign `String error = PR_INVALID_CODE` at 113.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtilTest.java
```diff
@@ -120,6 +120,19 @@ public void testCreatePullReq_errorCase1() throws Exception {
...
+        GitHub github = mock(GitHub.class);
+        GitHubUtil gitHubUtil = new GitHubUtil(github);
+        GHRepository origRepo = mock(GHRepository.class);
+        when(origRepo.getDefaultBranch()).thenReturn("master");
+        when(origRepo.createPullRequest(eq("title"), eq("owner:branch"), eq("master"), eq("body")))
+                .thenThrow(new IOException("{\"message\":\"Validation Failed\",\"errors\":[{\"resource\":\"PullRequest\",\"field\":\"head\",\"code\":\"invalid\"}],\"documentation_url\":\"https://developer.github.com/enterprise/2.6/v3/pulls/#create-a-pull-request\"}"));
+        GHRepository forkRepo = mock(GHRepository.class);
+        when(forkRepo.getOwnerName()).thenReturn("owner");
+        assertEquals(gitHubUtil.createPullReq(origRepo, "branch", forkRepo, "title", "body"), 1);
+        verify(origRepo, times(1)).createPullRequest(eq("title"), eq("owner:branch"), eq("master"), eq("body"));
```
cjneasbi, line 134: times(1) is the default for verify, you can remove the second parameter.

avimanyum:
> Hmm... Are there other scenarios that end up taking this path? My concern is that when you delete a fork, the API will lag by hours and may select a fork that is no longer there. Additionally, I believe this may have an effect on open PRs.

Hmm, in that case we might just want to make sure that we catch the exception by checking if the json field actually exists before accessing it, but not return 1 -> which is what triggers the delete of the fork.

avimanyum:
@cjneasbi Addressed your comments in this new PR - https://github.com/salesforce/dockerfile-image-update/pull/331

PR #326, opened 2022-03-08
Resetting the list of orgs to include or exclude in the search for each image

avimanyum:
While running the `all` subcommand, there is a Git search that happens for each image found in the tag store one after the other. While processing one of the images, if the hashmap orgsToIncludeInSearch is modified (this can happen one image returns more than 1000 search results. Refer to [this](https://github.com/salesforce/dockerfile-image-update/pull/276) for more details), the modified map persists and is carried over while performing the search for the next image in the tag store. This is a bug. 
This PR resets the hashmap before performing the search for each image.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
@@ -40,13 +40,6 @@ public class All implements ExecutableWithNamespace {
         Set<Map.Entry<String, JsonElement>> imageToTagStore =
                 this.dockerfileGitHubUtil.getGitHubJsonStore(ns.get(Constants.STORE)).parseStoreToImagesMap(dockerfileGitHubUtil, ns.get(Constants.STORE));
         Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
-        Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
-        if (ns.get(Constants.GIT_ORG) != null) {
-            // If there is a Git org specified, that needs to be included in the search query. In
-            // the orgsToIncludeInSearch a true value associated with an org name ensures that
-            // the org gets included in the search query.
-            orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
-        }
         List<ProcessingErrors> imagesThatCouldNotBeProcessed = new LinkedList<>();
         AtomicInteger numberOfImagesToProcess = new AtomicInteger();
         for (Map.Entry<String, JsonElement> imageToTag : imageToTagStore) {
@@ -59,7 +52,13 @@ public class All implements ExecutableWithNamespace {
                 GitForkBranch gitForkBranch = getGitForkBranch(image, tag, ns);
 
                 log.info("Finding Dockerfiles with the image name {}...", image);
-
+                Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
+                if (ns.get(Constants.GIT_ORG) != null) {
+                    // If there is a Git org specified, that needs to be included in the search query. In
+                    // the orgsToIncludeInSearch a true value associated with an org name ensures that
+                    // the org gets included in the search query.
+                    orgsToIncludeInSearch.put(ns.get(Constants.GIT_ORG), true);
+                }
                 Optional<List<PagedSearchIterable<GHContent>>> contentsWithImage =
                         this.dockerfileGitHubUtil.findFilesWithImage(image, orgsToIncludeInSearch, gitApiSearchLimit);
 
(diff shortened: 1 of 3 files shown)


afalko (approved):
It think it wants us to add a unit test for this case. If it is too difficult, I'm still OK merging.

cjneasbi:
So I think this change is solving the wrong problem.  The problem is a method in the call chain is programming by side effect instead of just modifying its return value.  If I understand correctly the map being cleared is passed [here](https://github.com/salesforce/dockerfile-image-update/blob/f9f36c2221206333bff929e6bb725562a849e722/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java#L63), to [here](https://github.com/salesforce/dockerfile-image-update/blob/f9f36c2221206333bff929e6bb725562a849e722/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L147), and then is inappropriately modified at [here](https://github.com/salesforce/dockerfile-image-update/blob/f9f36c2221206333bff929e6bb725562a849e722/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java#L175).  Thats the actual bug, `getSearchResultsExcludingOrgWithMostHits` should not be modifying its input parameter to which is programming by side effect.  Instead we should change the return value of `getSearchResultsExcludingOrgWithMostHits` to also return any additional information (like the updates to the input parameter) required by the callers.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -59,7 +52,13 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
                 GitForkBranch gitForkBranch = getGitForkBranch(image, tag, ns);
 
                 log.info("Finding Dockerfiles with the image name {}...", image);
-
+                Map<String, Boolean> orgsToIncludeInSearch = new HashMap<>();
```
cjneasbi, line 55: See comments about why I think this is solving the wrong problem.

cjneasbi (changes requested):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
avimanyum: The map is being modified as expected. The modification of the map happens only when we exclude orgs recursively to reduce the search space to get around the limit that Git has of returning only 1000 search results.
For every image that is being searched for from the image tag store, the map needs to start as an empty map. Only if the search of that image returns more than 1000 results does the map get modified - which is what we want. 
However, when that image is processed completely and the run proceeds to the next image, this map should again start as an empty map and not automatically exclude the orgs that were excluded during the search of some previous image from the tag store.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java
```diff
@@ -330,6 +330,7 @@ public void getSearchResultsExcludingOrgWithMostHits() throws Exception {
         Map<String, Boolean> orgsToIncludeOrExclude = new HashMap<>();
 
         assertEquals((dockerfileGitHubUtil.getSearchResultsExcludingOrgWithMostHits("image", contentsWithImage, orgsToIncludeOrExclude, 1000)).get().size(), 2);
+        assertEquals(orgsToIncludeOrExclude.size(), 0);
```
cjneasbi, line 333: Minor: We might want to add a small comment as to why we are checking for this condition.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -177,10 +177,11 @@ protected String getOrgNameWithMaximumHits(PagedSearchIterable<GHContent> files)
         Optional<List<PagedSearchIterable<GHContent>>> contentsForOrgWithMaximumHits;
         contentsForOrgWithMaximumHits = findFilesWithImage(image, orgsToInclude, gitApiSearchLimit);
 
-        orgsToExclude.put(orgWithMaximumHits, false);
-        log.info("Running search by excluding the orgs {}.", orgsToExclude.keySet().toString());
+        final Map<String, Boolean> orgsToExcludeFromSearch = new HashMap(orgsToExclude);
```
cjneasbi, line 180: You'll want to use `new HashMap<>(orgsToExclude)` otherwise you'll get type warnings during compilation.

cjneasbi (changes requested):

cjneasbi (approved):

afalko (approved):

PR #327, opened 2022-03-09
Label triggers are creating duplicate releases

justinharringa:
It seems like these triggers are kinda counterproductive. It'd be good
to keep the workflow_dispatch but it's unclear whether that will also
create a duplicate so I'll keep it for now.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,8 +1,6 @@
 name: Release Drafter
 
 on:
-  label:
-    types: [created, deleted]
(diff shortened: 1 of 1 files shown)


avimanyum:
Will this stop the release on cases when we actually want the release to happen?

justinharringa:
> Will this stop the release in cases when we actually want the release to happen?

This is orthogonal to the actual releases. See https://github.com/release-drafter/release-drafter for more information.

justinharringa:
No. This is simply drafting release notes

PR #331, opened 2022-03-14
Use stream to look for error message prefixes

avimanyum:
This PR creates an Array of potential error messages and streams through them to check for the existence of one. This way we will not have to modify the conditional check even if there are more error prefixes we add in the future.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -11,6 +11,7 @@ package com.salesforce.dockerfileimageupdate.utils;
 import com.google.common.collect.Multimap;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 import org.kohsuke.github.*;
 import org.slf4j.Logger;
@@ -21,6 +22,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -110,18 +112,18 @@ public int createPullReq(GHRepository origRepo, String branch,
...
-                error = PR_INVALID_CODE;
+            String error = PR_INVALID_CODE;
+            JsonObject errorMessage = errorJson.get(0).getAsJsonObject();
+            if (errorMessage.has("message")) {
+                error = errorMessage.get("message").getAsString();
             }
             log.info("error: {}", error);
+            String finalError = error;
+            List<String> errorPrefixes = Arrays.asList(PR_INVALID_CODE, "No commits between");
             if (error.startsWith("A pull request already exists")) {
```
cjneasbi, line 123: Not sure why you need `finalError` instead of just using `error` but you should probably use either one or the other in both if statements.

cjneasbi (approved):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -110,18 +112,18 @@ public int createPullReq(GHRepository origRepo, String branch,
...
             JsonElement root = JsonParser.parseString(e.getMessage());
             JsonArray errorJson = root.getAsJsonObject().get("errors").getAsJsonArray();
-            String error = "";
-            if (errorJson.get(0).getAsJsonObject().has("message")) {
-                error = errorJson.get(0).getAsJsonObject().get("message").getAsString();
-            } else {
-                // This case usually happens when the PR head has invalid data. Deleting the forked repo resolves it.
-                error = PR_INVALID_CODE;
+            String error = PR_INVALID_CODE;
+            JsonObject errorMessage = errorJson.get(0).getAsJsonObject();
```
afalko, line 116: Are there ever more than 1 error in this array? Any reason not to `foreach` on them?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
avimanyum: When I tried using the variable `error` I got the following warning on IDE
`Variable used in lambda expression should be final or effectively final`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
avimanyum: Not sure about that. This piece of code has been there for a while.

afalko (approved):

PR #333, opened 2022-03-28
[WIP] Initial interface for storage

afalko:

PR #334, opened 2022-03-28
Enable S3 backed image tag store

avimanyum:
This PR enables DFIU to read from and write to an S3 backed image tag store.


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+public class S3Store {
```
afalko, line 20: We need to create a common interface for this and the older Git based store. See what I have in https://github.com/salesforce/dockerfile-image-update/pull/333

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -47,9 +48,10 @@ public static void main(String[] args)
             System.exit(1);
         Class<?> runClass = loadCommand(allClasses, ns.get(Constants.COMMAND));
         DockerfileGitHubUtil dockerfileGitHubUtil = initializeDockerfileGithubUtil(ns.get(Constants.GIT_API));
+        DockerfileS3Util dockerfileS3Util = initializeDockerfileS3Util();
```
afalko, line 51: No need to initialize both. We shouldn't need to create a "util" for S3 and keep all methods with the S3Store you created below.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -47,9 +48,10 @@ public static void main(String[] args)
...
 
         /* Execute given command. */
-        ((ExecutableWithNamespace)runClass.newInstance()).execute(ns, dockerfileGitHubUtil);
+        ((ExecutableWithNamespace)runClass.newInstance()).execute(ns, dockerfileGitHubUtil, dockerfileS3Util);
```
afalko, line 54: We can avoid this too if we detect the store type and create the right instance of our Store interface.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /* The store link should be a bucket name on S3. */
+    public void updateS3Store(String img, String tag) throws IOException {
+        if (store == null) {
+            log.info("Image tag store cannot be null. Skipping store update...");
+            return;
+        }
+        log.info("Updating store: {} with image: {} tag: {}...", store, img, tag);
+        if (!s3BucketUtil.getS3Client().doesBucketExistV2(store)) {
+            log.error("The S3 bucket with name {} does not exist. Cannot proceed.", store);
```
afalko, line 38: We should throw exception. I don't think this error can be ignored.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,55 @@
...
+        }
+        String key = convertImageStringToS3ObjectKey(img);
+        s3BucketUtil.getS3Client().putObject(store, key, tag);
+    }
+
+    protected String convertImageStringToS3ObjectKey(String img) {
+        return img.replace('/', ';');
```
afalko, line 45: Make `;` a constant called `S3_FILE_KEY_PATH_DELIMITER` or something like that.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public String convertS3ObjectKeyToImageString(String key) {
```
afalko, line 48: Reuse delimiter constant here

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -8,20 +8,26 @@
...
-import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
-import com.salesforce.dockerfileimageupdate.utils.ProcessingErrors;
-import com.salesforce.dockerfileimageupdate.utils.PullRequests;
+import com.salesforce.dockerfileimageupdate.utils.*;
 import net.sourceforge.argparse4j.inf.Namespace;
 import org.kohsuke.github.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.swing.text.html.Option;
```
afalko, line 30: This looks like an accidental import

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -30,15 +36,16 @@
         requiredParams = {Constants.STORE})
 public class All implements ExecutableWithNamespace {
     private static final Logger log = LoggerFactory.getLogger(All.class);
+    private static final String s3Prefix = "s3://";
```
afalko, line 39: Curious if the S3 SDK has a constant for this already

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -49,39 +56,73 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
-                            pullRequests.prepareToCreate(ns, pullRequestSender,
-                                    pagedSearchIterable, gitForkBranch, dockerfileGitHubUtil);
-                        } catch (IOException e) {
-                            log.error("Could not send pull request for image {}.", image);
-                            processErrors(image, tag, e, imagesThatCouldNotBeProcessed);
-                        }
-                    });
+        String store = ns.get(Constants.STORE);
+        Optional<Exception> failureMessage;
+        if (store.startsWith(s3Prefix)) {
```
afalko, line 61: No need for an if statement. We should be calling our Store interface and have all this S3 stuff in the S3 implementation. Roughly speaking your code will look like this:

URI storeUri = new URI(store);
ImageTagStore imageTagStore;
switch (storeUri.getScheme()) {
  case("s3"):
        imageTagStore = new S3Store(...);
        break;
  default:
        imageTagStore = ... /* initialize like we did before */
}

imageTagStore.parseStoreToImagesMap(...)
  
}

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Parent.java
```diff
@@ -33,18 +32,29 @@
...
 
     DockerfileGitHubUtil dockerfileGitHubUtil;
+    DockerfileS3Util dockerfileS3Util;
 
     @Override
-    public void execute(final Namespace ns, DockerfileGitHubUtil dockerfileGitHubUtil)
+    public void execute(final Namespace ns, DockerfileGitHubUtil dockerfileGitHubUtil, DockerfileS3Util dockerfileS3Util)
             throws IOException, InterruptedException {
         loadDockerfileGithubUtil(dockerfileGitHubUtil);
+        loadDockerfileS3Util(dockerfileS3Util);
```
afalko, line 44: Same feedback here as I add for `All`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -115,4 +115,15 @@ protected String getAndModifyJsonString(JsonElement json, String img, String tag
         JsonElement imagesJson = json.getAsJsonObject().get("images");
         return imagesJson.getAsJsonObject().entrySet();
     }
+
+    public HashMap<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) throws IOException, InterruptedException {
```
afalko, line 119:
```suggestion
    public Map<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) throws IOException, InterruptedException {
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -115,4 +115,15 @@ protected String getAndModifyJsonString(JsonElement json, String img, String tag
...
+        HashMap<String, String> imageNameWithTag = new HashMap<>();
```
afalko, line 120:
```suggestion
        Map<String, String> imageNameWithTag = new HashMap<>();
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/ImageTagStore.java
```diff
@@ -0,0 +1,11 @@
+package com.salesforce.dockerfileimageupdate.storage;
+
+import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
+
+import java.io.IOException;
+import java.util.HashMap;
+
+public interface ImageTagStore {
+    void updateStore(String img, String tag) throws IOException;
+    HashMap<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) throws IOException, InterruptedException;
```
afalko, line 10:
```suggestion
    Map<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) throws IOException, InterruptedException;
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/ImageTagStore.java
```diff
@@ -0,0 +1,11 @@
+package com.salesforce.dockerfileimageupdate.storage;
+
+import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
+
+import java.io.IOException;
+import java.util.HashMap;
+
+public interface ImageTagStore {
+    void updateStore(String img, String tag) throws IOException;
```
afalko, line 9: We should add some javadoc to these interface methods.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /* The store link should be a bucket name on S3. */
+    public void updateStore(String img, String tag) throws IOException {
+        if (store == null) {
+            log.info("Image tag store cannot be null. Skipping store update...");
+            return;
+        }
+        log.info("Updating store: {} with image: {} tag: {}...", store, img, tag);
+        if (!s3.doesBucketExistV2(store)) {
```
afalko, line 32: Can we make this a positive condition and do the throw in `else`?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,68 @@
...
+            throw new IOException(String.format("The S3 bucket with name %s does not exist. Cannot proceed.", store));
+        }
+        String key = convertImageStringToS3ObjectKey(img);
+        s3.putObject(store, key, tag);
+    }
+
+    public HashMap<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) {
+        HashMap<String, String> imageNameWithTag = new HashMap<>();
```
afalko, line 40:
```suggestion
        Map<String, String> imageNameWithTag = new HashMap<>();
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -49,37 +51,61 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
-                GitForkBranch gitForkBranch = getGitForkBranch(image, tag, ns);
-
-                log.info("Finding Dockerfiles with the image name {}...", image);
+        Optional<Exception> failureMessage;
+        String store = ns.get(Constants.STORE);
+        URI storeUri = new URI(store);
+        ImageTagStore imageTagStore;
+        switch (storeUri.getScheme()) {
+            case (s3Prefix):
+                log.info("The underlying data store is an S3 bucket.");
```
afalko, line 60:
```suggestion
                log.info("The underlying data store is S3.");
```

afalko: We might have a path within a bucket

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -49,37 +51,61 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+                AmazonS3 s3 = AmazonS3ClientBuilder.standard().withRegion(Regions.DEFAULT_REGION).build();
+                imageTagStore = new S3Store(s3, store);
+                break;
+            default:
+                log.info("The underlying data store is a Git repository.");
+                imageTagStore = this.dockerfileGitHubUtil.getGitHubJsonStore(store);
+        }
+        HashMap<String, String> imageNameWithTag = imageTagStore.getStoreContent(dockerfileGitHubUtil, store);
```
afalko, line 68:
```suggestion
        Map<String, String> imageNameWithTag = imageTagStore.getStoreContent(dockerfileGitHubUtil, store);
```

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
```diff
@@ -19,6 +19,7 @@
 import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import com.sun.xml.internal.bind.v2.TODO;
```
afalko, line 22: Accidental import?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,68 @@
...
+        ListObjectsV2Result result = getS3Objects();
+        List<S3ObjectSummary> objects = result.getObjectSummaries();
+        for (S3ObjectSummary os : objects) {
+            String key = os.getKey();
+            String image = convertS3ObjectKeyToImageString(key);
+            S3Object o = getS3Object(store, key);
+            String tag = o.getObjectContent().toString();
+            imageNameWithTag.put(image, tag);
+        }
+        return imageNameWithTag;
```
afalko, line 50: I'm assuming we'll sort by last accessed time in a future PR?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public HashMap<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) {
+        HashMap<String, String> imageNameWithTag = new HashMap<>();
+        ListObjectsV2Result result = getS3Objects();
+        List<S3ObjectSummary> objects = result.getObjectSummaries();
+        for (S3ObjectSummary os : objects) {
+            String key = os.getKey();
+            String image = convertS3ObjectKeyToImageString(key);
+            S3Object o = getS3Object(store, key);
```
afalko, line 46: We might get throttled by S3 here, so we might need to inject a short sleep in this loop.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -9,11 +9,12 @@
...
 
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.google.common.reflect.ClassPath;
 import com.salesforce.dockerfileimageupdate.subcommands.ExecutableWithNamespace;
-import com.salesforce.dockerfileimageupdate.utils.Constants;
-import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
-import com.salesforce.dockerfileimageupdate.utils.GitHubUtil;
+import com.salesforce.dockerfileimageupdate.utils.*;
```
cjneasbi, line 17: Unless you are using the whole package stay away from star imports, see https://www.javadude.com/posts/20040522-import-on-demand-is-evil/

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/ImageTagStore.java
```diff
@@ -0,0 +1,26 @@
+package com.salesforce.dockerfileimageupdate.storage;
+
+import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
+
+import java.io.IOException;
+import java.util.Map;
+
+public interface ImageTagStore {
```
cjneasbi, line 8: Add a javadoc here describing what this interface is defining.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/ImageTagStore.java
```diff
@@ -0,0 +1,26 @@
...
+    void updateStore(String img, String tag) throws IOException;
+
+    /**
+     * This method gets the content of the image tag store.
+     *
+     * @param dockerfileGitHubUtil the dockerfileGitHubUtil object that is used to interact with an underlying Git repo.
+     * @param storeName the name of the store whose content needs to be fetched.
+     * @return A Map of image name to image version.
+     */
+    Map<String, String> getStoreContent(DockerfileGitHubUtil dockerfileGitHubUtil, String storeName) throws IOException, InterruptedException;
```
cjneasbi, line 25: Instead of `Map<String, String>` just create a new class to store the data you want.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+import com.salesforce.dockerfileimageupdate.utils.DockerfileGitHubUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+public class S3Store implements ImageTagStore {
```
cjneasbi, line 16: Rename to `S3BackedImageTagStore` and add a javadocs

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final Logger log = LoggerFactory.getLogger(S3Store.class);
+    private static final Character S3_FILE_KEY_PATH_DELIMITER = ';';
```
cjneasbi, line 18: See https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html .  I don't think `;` is a safe character of object names.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.*;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+public class S3Store implements ImageTagStore {
+    private static final Logger log = LoggerFactory.getLogger(S3Store.class);
```
cjneasbi, line 17: Java convention is constants are all upper case.  A java constant is anything that is static and final.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+    private final String store;
+
+    public S3Store(AmazonS3 s3, String store) {
+        this.s3 = s3;
+        this.store = store;
+    }
+
+    /* The store link should be a bucket name on S3. */
+    public void updateStore(String img, String tag) throws IOException {
+        if (store == null) {
```
cjneasbi, line 30: We should not accept null as a valid value.  I would validate that the parameters to the constructor are non-null.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.stream.Collectors;
+
+public class S3Store implements ImageTagStore {
+    private static final Logger log = LoggerFactory.getLogger(S3Store.class);
+    private static final Character S3_FILE_KEY_PATH_DELIMITER = ';';
+    public static final String s3Prefix = "s3";
+    private final AmazonS3 s3;
+    private final String store;
+
+    public S3Store(AmazonS3 s3, String store) {
```
cjneasbi, line 23: Method parameters should be final to prevent mutating the reference

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+        this.store = store;
+    }
+
+    /* The store link should be a bucket name on S3. */
+    public void updateStore(String img, String tag) throws IOException {
+        if (store == null) {
+            log.info("Image tag store cannot be null. Skipping store update...");
+            return;
+        }
+        log.info("Updating store: {} with image: {} tag: {}...", store, img, tag);
```
cjneasbi, line 34: Lets use `debug` instead of `info` as `info` is usually the default log level.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+        if (s3.doesBucketExistV2(store)) {
+            String key = convertImageStringToS3ObjectKey(img);
+            s3.putObject(store, key, tag);
+        } else {
+            throw new IOException(String.format("The S3 bucket with name %s does not exist. Cannot proceed.", store));
```
cjneasbi, line 39: Java checked exceptions are stupid,  here is why http://literatejava.com/exceptions/checked-exceptions-javas-biggest-mistake/ .  In this case just throw and unchecked exception.  https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html is perfect for this case.

Also don't use string.format in Java.  It's literally the slowest formatting you can have, see https://redfin.engineering/java-string-concatenation-which-way-is-best-8f590a7d22a8

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+                        Map.Entry::getValue,
+                        (oldValue, newValue) -> oldValue, LinkedHashMap::new));
+        for (Map.Entry<String, Date> set : sortedResult.entrySet()) {
+            String key = set.getKey();
+            String image = convertS3ObjectKeyToImageString(key);
+            S3Object o = getS3Object(store, image);
+            String tag = o.getObjectContent().toString();
+            imageNameWithTagSortedByAccessDate.put(image, tag);
+            //Adding a wait to avoid S3 throttling the requests
+            waitFor(TimeUnit.SECONDS.toMillis(1));
```
cjneasbi, line 74: Not a good idea.  You typically want to react to back pressure instead of trying to time your requests.  The AWS sdk already had backoff and retry mechanisms built in so I would just use them,  see https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        return imageNameWithTagSortedByAccessDate;
+    }
+
+    protected void waitFor(long millis) throws InterruptedException {
+        Thread.sleep(millis);
```
cjneasbi, line 80: Unless you are periodically pausing a background thread of some type of server application you shouldn't use Thread.sleep.  Fortunately in this case you don't need it.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -35,10 +38,8 @@
 
     @Override
     public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGitHubUtil)
```
cjneasbi, line 40: This method has quite a bit of logic in it.  Lets break it up into some smaller helper methods.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -49,37 +50,61 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
+        ImageTagStore imageTagStore;
+        switch (storeUri.getScheme()) {
+            case (S3Store.s3Prefix):
+                log.info("The underlying data store is S3.");
+                AmazonS3 s3 = AmazonS3ClientBuilder.standard().withRegion(Regions.DEFAULT_REGION).build();
+                imageTagStore = new S3Store(s3, store);
+                break;
+            default:
+                log.info("The underlying data store is a Git repository.");
+                imageTagStore = this.dockerfileGitHubUtil.getGitHubJsonStore(store);
```
cjneasbi, line 65: This is a good use case for an Enum, create one value for S3 and value for Git.  Create a static method called `fromStoreUri` that produces an enum based upon the uri parameter.  Each enum value has a `getStore` method that returns the correct tag store implementation.

cjneasbi (changes requested):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
afalko: Great find! I recommend going with `.`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
afalko: I think in this case, we might want to see what is going on, no? Debug would be for something much more detailed that we don't want to have polluting out logs. I think this log line is more signal that noise, but feel free to convince me that is not the case.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
afalko: For String.format, I'd rather have readability - I don't think its much of a perf hit all things considered. Good point though.

cjneasbi: I would agree on readability if there are a bunch of things being replaced.  But this is a single string value so

`"The S3 bucket with name " + store + " does not exist. Cannot proceed."`

is just as readable as

`String.format("The S3 bucket with name %s does not exist. Cannot proceed.", store)`

IMHO, Also it has fewer method calls and characters.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
cjneasbi: "I think in this case, we might want to see what is going on, no?"  Only if something is failing.  If it not we will never look at these logs.  If it is we can conditionally enable debug logging and re-run.  If we want this information on every run I would advise building a file containing information about what was touched and what failed and return that as the output of the command.  Sort of like our published artifacts files.  DFIU also runs faster if it prints less.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -100,10 +120,9 @@ protected void printSummary(List<ProcessingErrors> imagesThatCouldNotBeProcessed
         }
     }
 
-    protected void processErrors(String image, String tag, Exception e, List<ProcessingErrors> imagesThatCouldNotBeProcessed) {
+    protected ProcessingErrors processErrorMessages(String image, String tag, Optional<Exception> failure) {
         String imageNameAndTag = image + ":" + tag;
-        ProcessingErrors processingErrors = new ProcessingErrors(imageNameAndTag, Optional.of(e));
-        imagesThatCouldNotBeProcessed.add(processingErrors);
+        return new ProcessingErrors(imageNameAndTag, failure);
```
cjneasbi, line 125: Why not just have three values in the class instead of two?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -35,10 +38,18 @@
...
-            throws IOException, InterruptedException {
+            throws IOException, InterruptedException, URISyntaxException {
         loadDockerfileGithubUtil(dockerfileGitHubUtil);
-        Set<Map.Entry<String, JsonElement>> imageToTagStore =
-                this.dockerfileGitHubUtil.getGitHubJsonStore(ns.get(Constants.STORE)).parseStoreToImagesMap(dockerfileGitHubUtil, ns.get(Constants.STORE));
+        String store = ns.get(Constants.STORE);
+        ImageStoreUtil imageStoreUtil = getImageStoreUtil();
+        ImageTagStore imageTagStore = imageStoreUtil.initializeImageTagStore(this.dockerfileGitHubUtil, store);
+        List<ImageTagStoreContent> imageNamesWithTag = imageTagStore.getStoreContent(dockerfileGitHubUtil, store);
+        AtomicInteger numberOfImagesToProcess = new AtomicInteger(imageNamesWithTag.size());
```
cjneasbi, line 47: I'm not sure why we need an AtomicInteger here?  This code does not appear multi-threaded an integer assignment in Java is already atomic.  See https://stackoverflow.com/questions/1006655/are-java-primitive-ints-atomic-by-design-or-by-accident .  If we don't need it just use `int`.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/ImageStoreUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * ImageStore is the main entity we'll be using to initialize the image tag store
+ * @author amukhopadhyay
+ */
+public class ImageStoreUtil {
```
cjneasbi, line 16: This looks like a utility class as it doesn't store instance state.  In that case, make all of the methods static, make this class final and add the following constructor:

```
private ImageStoreUtil() {}
```
This prevents the creation of instances of this class.

cjneasbi (changes requested):

dockerfile-image-update/pom.xml
```diff
@@ -122,6 +122,11 @@
             <artifactId>github-api</artifactId>
             <version>1.303</version>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-s3</artifactId>
+            <version>1.11.641</version>
```
cjneasbi, line 128: See https://github.com/aws/aws-sdk-java-v2/#using-the-sdk .  We should probably use the latest version of the SDK unless we have good reason not to.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -13,8 +13,9 @@
 import java.io.InputStreamReader;
 import java.nio.file.Paths;
 import java.util.*;
+import java.util.stream.Collectors;
```
cjneasbi, line 16: Minor:  A nice feature in java reduce the terseness of code is a static import

instead of 
```
import java.util.stream.Collectors;

...

Collectors.toList()
```

You can do

```
import static java.util.stream.Collectors.toList;

...

toList()
```

This is a good pattern to follow for utility class methods especially in tests.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/ImageStoreType.java
```diff
@@ -0,0 +1,34 @@
...
+        @Override
+        public ImageTagStore getStore(DockerfileGitHubUtil dockerfileGitHubUtil, String store) {
+            AmazonS3 s3 = AmazonS3ClientBuilder.standard().withRegion(Regions.DEFAULT_REGION).build();
+            return new S3BackedImageTagStore(s3, store);
+        }
+    },
+    GIT{
+        @Override
+        public ImageTagStore getStore(DockerfileGitHubUtil dockerfileGitHubUtil, String store) {
+            return dockerfileGitHubUtil.getGitHubJsonStore(store);
```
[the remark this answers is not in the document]
cjneasbi: Yeah looking at that code it seems like its being using in multiple operations which is why its initialized out side of the is class.  If they are cheep to create have two and remove the unneeded parameter.  If they are not then I would comment that the parameter is intended to be ignored in the case of S3 or add a helper method `getStore(String store)` that calls the currently existing `getStore(DockerfileGitHubUtil dockerfileGitHubUtil, String store)` but passes null as the first parameter.  This way callers don't have to worry about that first parameter at all in the S3 case.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -37,8 +39,21 @@
...
+        ImageStoreUtil imageStoreUtil = getImageStoreUtil();
+        try {
+            ImageTagStore imageTagStore = imageStoreUtil.initializeImageTagStore(this.dockerfileGitHubUtil, store);
+            List<ImageTagStoreContent> imageNamesWithTag = imageTagStore.getStoreContent(dockerfileGitHubUtil, store);
+            Integer numberOfImagesToProcess = imageNamesWithTag.size();
+            List<ProcessingErrors> imagesThatCouldNotBeProcessed = processImagesWithTag(ns, imageNamesWithTag);
+            printSummary(imagesThatCouldNotBeProcessed, numberOfImagesToProcess);
+        } catch (URISyntaxException e) {
+            log.error("Could not initialize the Image tage store. Exception: ", e.getMessage());
+        }
```
cjneasbi, line 52: This is good.  I but I didn't intend for the checked exception to be caught and swallowed here.  I should have been more clear.  We should catch the checked exception and yest we can log the exception.  But we should wrap it in an unchecked exception and throw that.  Then we aren't handling exceptions prematurely but we also aren't throwing checked exceptions that break the method signature.  Ideally we create `DFIU` specific unchecked exceptions and throw those so we have abstracted our dependencies from our customers.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -47,63 +62,70 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
...
-                        }
-                    });
+        for (ImageTagStoreContent content : imageNamesWithTag) {
+            String image = content.getImageName();
+            String tag = content.getTag();
+            failureMessage = processImageWithTag(image, tag, ns, orgsToIncludeInSearch, gitApiSearchLimit);
+            if (failureMessage.isPresent()) {
+                ProcessingErrors processingErrors = processErrorMessages(image, tag, failureMessage);
+                imagesThatCouldNotBeProcessed.add(processingErrors);
+            }
```
cjneasbi, line 74: This is a good place for `ifPresent`, https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-

```
failureMessage.ifPresent(message -> imagesThatCouldNotBeProcessed.add(processErrorMessages(image, tag, message)));
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/ImageStoreUtil.java
cjneasbi: "This is being initialized in the All and Parent class."  With what?  Basically if you are doing the following in Java
```
ImageStoreUtil util = new ImageStoreUtil()
util.initializeImageTagStore();
```
And `ImageStoreUtil` maintains no instance (i.e non-static) variables then this is equivalent but less resource intensive
```
ImageStoreUtil.initializeImageTagStore();
```
You just need to make initializeImageTagStore static because its a pure function.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ChildTest.java
```diff
@@ -46,38 +46,42 @@
...
+        ImageTagStore imageTagStore = mock(ImageTagStore.class);
+        when(child.getImageStoreUtil()).thenReturn(imageStoreUtil);
+        when(imageStoreUtil.initializeImageTagStore(dockerfileGitHubUtil, "test")).thenReturn(imageTagStore);
+        doNothing().when(dockerfileGitHubUtil).createPullReq(any(), anyString(), any(), any());
 
         child.execute(ns, dockerfileGitHubUtil);
 
-        Mockito.verify(dockerfileGitHubUtil, atLeastOnce())
-                .createPullReq(Mockito.any(), anyString(), Mockito.any(), any());
+        verify(dockerfileGitHubUtil, atLeastOnce())
```
cjneasbi, line 63: Why at least once and not exactly once?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/ImageStoreUtil.java
afalko: @cjneasbi isn't it going to be hard to mock out static methods? We generally follow the object initialization pattern so we can mock the while utility out when not unit testing the util itself.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3Store.java
avimanyum: @afalko @cjneasbi Using `.` will make it difficult to regenerate the correct image name when we read the content from the S3 bucket since the filename for an image `dva-registry.internal.salesforce.com/sfci/sfci-dfiu-test/dfiu-test-multiple-from` will now look like this 
`dva-registry.internal.salesforce.com.sfci.sfci-dfiu-test.dfiu-test-multiple-from`. I suggest using `!` as the delimiter since I have not seen that character being used in org, repo or image name. Thoughts?

afalko: Good point! `!` seems good to me

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/ImageStoreUtil.java
cjneasbi: In cases where you are trying to call a static method replace this with a Lambda parameter and that static method call is no longer static.

```
public SomeClass{
public static void doAThing();
}

...

public myMethod(){
  SomeClass.doAThing()
}
```

becomes

```
public SomeClass{
public static void doAThing();
}

...

public myMethod(){
  myMethod(SomeClass::doAThing)
}

myMethod(Supplier<Void> thingDoer){
  thingDoer.get()
}

``` 

Then you test `myMethod(Supplier<Void> thingDoer)`

```
@Test
public void myTest()
{
  myMethod(() -> null)
}
```
 
Now you don't need Mockito to mock your static logic

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/S3BackedImageTagStore.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        return imageNameWithTagSortedByAccessDate;
+    }
+
+    private String getTagValueFromObject(S3Object o) {
+        String tag = "";
+        try (S3ObjectInputStream is = o.getObjectContent()) {
+            tag = CharStreams.toString(new InputStreamReader(is, Charsets.UTF_8));
+        } catch (IOException e) {
+            log.error("Could not get the tag value. Exception: ", e.getMessage());
```
afalko, line 101: Is there is a reason we are not bubbling this exception up and swallowing it here instead?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -34,11 +35,21 @@ public class All implements ExecutableWithNamespace {
...
-                this.dockerfileGitHubUtil.getGitHubJsonStore(ns.get(Constants.STORE)).parseStoreToImagesMap(dockerfileGitHubUtil, ns.get(Constants.STORE));
+        String store = ns.get(Constants.STORE);
+        try {
+            ImageTagStore imageTagStore = ImageStoreUtil.initializeImageTagStore(this.dockerfileGitHubUtil, store);
+            List<ImageTagStoreContent> imageNamesWithTag = imageTagStore.getStoreContent(dockerfileGitHubUtil, store);
+            Integer numberOfImagesToProcess = imageNamesWithTag.size();
+            List<ProcessingErrors> imagesThatCouldNotBeProcessed = processImagesWithTag(ns, imageNamesWithTag);
+            printSummary(imagesThatCouldNotBeProcessed, numberOfImagesToProcess);
+        } catch (Exception e) {
+            log.error("Encountered issues while initializing the image tag store or getting its contents. Exception: ", e);
```
afalko, line 48: Isn't this fatal? Why not throw? Seems like we should exit with non-zero exit code if this happens

afalko (approved):

afalko:
Merged due to Chris being OOTO - we can do a separate PR for any remaining items.ng due to Chris being OOTO - we can do a separate PR for any remaining items.

PR #533, opened 2022-08-31
updated tag store read/write to blob

jeetchoudhary:
Test Cases:
new store created: 
    https://git.****.****.com/jitender-kumar/sfci-test/blob/master/store_new.json

store update with sized less < 1 MB 
    https://git.****.****.com/jitender-kumar/sfci-test/blob/master/store_new.json

store updated with size > 1MB
    https://git.****.****.com/jitender-kumar/sfci-test/blob/master/store.json

Benchmarking for reads from the new API:
image


jeetchoudhary:
Background : [github-api] GHRepository.getFileContent doesn't support reading/writing a file with size more > 1 MB

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -43,31 +49,51 @@ public void updateStore(String img, String tag) throws IOException {
...
+    private void initializeTagStoreIfRequired(GHRepository repo, String path) throws IOException {
         try {
             repo.getFileContent(path);
-        } catch (IOException e) {
+        } catch (FileNotFoundException e) {
+            log.info("Image tag store {} not found. Creating a new image tag store", path);
             repo.createContent().content("").message("initializing store").path(path).commit();
+        } catch (IOException ex) {
+            if (ex.getMessage().contains("too_large")) {
+                log.info("Image tag store {} is already initialized and have size more than 1 MB"
```
afalko, line 60:
```suggestion
                log.info("Image tag store {} is already initialized and has size more than 1 MB"
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -43,31 +49,51 @@ public void updateStore(String img, String tag) throws IOException {
...
+                        , path);
+            } else {
+                throw ex;
```
afalko, line 63: I don't think we threw before - just want to make sure this is intended.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -43,31 +49,51 @@ public void updateStore(String img, String tag) throws IOException {
...
-                    String.format("Updated image %s with tag %s.\n@rev none@", img, tag), repo.getDefaultBranch());
+    protected void updateStoreOnGithub(GHRepository repo, String path, String img, String tag) throws IOException {
+        initializeTagStoreIfRequired(repo, path);
+
+        GHCommit commit = repo.getCommit(repo.getDefaultBranch());
+        GHBlob blob = commit.getTree().getEntry(path).asBlob();
+        JsonElement json = null;
+
+        try {
+            String text = IOUtils.toString(blob.read(), StandardCharsets.UTF_8.name());
```
afalko, line 76: I rather not add an extra dependency on IOUtils. You should be able to use the language builtin here:
```
new String(blob.read().readAllBytes(), StandardCharsets.UTF_8)
```

afalko: Also, your blob.read() returns a Closable, so you need to add the try with resources back. I recommend something like this: 
```
String text;
try (InputStream stream = blob.read()) {
    text = new String(blob.read().readAllBytes(), StandardCharsets.UTF_8);
}

```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -43,31 +49,51 @@ public void updateStore(String img, String tag) throws IOException {
...
+            json = JsonParser.parseString(text);
+        } catch (JsonParseException e) {
+            log.warn("Not a JSON format store. Clearing and rewriting as JSON...");
```
afalko, line 79:
```suggestion
            log.warn("Not a JSON formated store. Clearing and rewriting as JSON...");
```

dockerfile-image-update/pom.xml
```diff
@@ -139,6 +139,12 @@
             <version>3.12.0</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
afalko, line 142: No need, see comment below

afalko (changes requested):
Thanks for the update! Small adjustment and a few typos in messages to resolve.

afalko (approved):

PR #543, opened 2022-09-01
Read BLOB for all sub-command

jeetchoudhary:


jeetchoudhary:
The idea is to read blob only for the tag store since the size of tag store can go beyond 1 MB, for normal operations we should still read the content API as it's little more efficient then reading blob

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
```diff
@@ -160,10 +159,13 @@ public Set<Map.Entry<String, JsonElement>> parseStoreToImagesMap(DockerfileGitHu
         }
 
         JsonElement imagesJson = json.getAsJsonObject().get("images");
+        log.info(imagesJson.toString());
+        System.exit(0);
```
afalko, line 163: Is this a remnant of debugging?

jeetchoudhary: removed this in the last commit

afalko (approved):
LGTM, thanks!

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/storage/GitHubJsonStore.java
jeetchoudhary: that didn't get pushed because of the squashed commit. just did that

PR #588, opened 2022-09-28
Use rateLimiter for PR throttling

tarishij17:
Provide a way to throttle number of PRs DFIU cuts over a period of time


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,30 @@
...
+public class RateLimiter {
+
+    private int rateLimit;
+    private Bucket bucket;
+
+    public RateLimiter() {
+
+        this.rateLimit = Constants.RATE_LIMIT;
+        //refill the bucket at the end of every hour with 'rateLimit' tokens, not exceeding the max capacity
+        Refill refill = Refill.intervally(rateLimit, Duration.ofHours(1));
```
cjneasbi, line 18: Is there a reason why the fill rate is not configurable?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,30 @@
+package com.salesforce.dockerfileimageupdate.utils;
+
+import io.github.bucket4j.Bandwidth;
+import io.github.bucket4j.Bucket;
+import io.github.bucket4j.Bucket4j;
+import io.github.bucket4j.Refill;
+import java.time.Duration;
+
+public class RateLimiter {
```
cjneasbi, line 9: We should have javadocs that describes what this class is an what its used for?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -52,6 +53,8 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
 
     protected List<ProcessingErrors> processImagesWithTag(Namespace ns, List<ImageTagStoreContent> imageNamesWithTag) {
         Integer gitApiSearchLimit = ns.get(Constants.GIT_API_SEARCH_LIMIT);
+        RateLimiter rateLimiter = getRateLimiter();
```
cjneasbi, line 56: We should probably create this as at the `execute` method level and pass this as a parameter to processImageWithTag.  We should create these utilities as close to the entry point of this class as possible.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/Child.java
```diff
@@ -63,15 +64,21 @@ public void execute(final Namespace ns, final DockerfileGitHubUtil dockerfileGit
 
         dockerfileGitHubUtil.createOrUpdateForkBranchToParentDefault(repo, fork, gitForkBranch);
 
+        RateLimiter rateLimiter = getRateLimiter();
```
cjneasbi, line 67: Since every command is going to need a rate limiter can we create the rate limiter outside of the commands and pass them to the command instance.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
```diff
@@ -37,5 +37,7 @@ private Constants() {
     public static final String GIT_API_SEARCH_LIMIT = "ghapisearchlimit";
     public static final String SKIP_PR_CREATION = "skipprcreation";
     public static final String IGNORE_IMAGE_STRING = "x";
+    //max number of PRs to send per hour
+    public static final int RATE_LIMIT = 30;
```
cjneasbi, line 41: Lets rename to `DEFAULT_RATE_LIMIT`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -358,7 +359,16 @@ public GitHubJsonStore getGitHubJsonStore(String store) {
...
                               String branch, GHRepository forkRepo,
-                              PullRequestInfo pullRequestInfo) throws InterruptedException, IOException {
+                              PullRequestInfo pullRequestInfo,
+                              RateLimiter rateLimiter) throws InterruptedException, IOException {
+
+        while (!rateLimiter.checkLimit()) {
+            //we have exceeded the limit and have no more tokens in the bucket
+            //wait for 10s before next try
+            log.info("We have exceeded the limit of PR creation per hour.. sleeping for 10s");
+            Thread.sleep(10000);
```
cjneasbi, line 369: We don't need this.  The rate limiter calls are blocking if a token is not available.  We should try to acquire a token and block until one is available instead of checking and then sleeping for 10 seconds just to sleep again.  See https://bucket4j.com/8.1.0/toc.html#using-bucket-as-throttler and https://bucket4j.com/8.1.0/toc.html#limiting-the-rate-of-access-to-rest-api as examples.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/AllTest.java
```diff
@@ -213,7 +221,7 @@ public void testAllCommandSkipsSendingPRsIfPRCreationRaisesException() throws Ex
         verify(all, times(1)).getPullRequestSender(dockerfileGitHubUtil, ns);
         verify(all, times(1)).getPullRequests();
         verify(pullRequests, times(1)).prepareToCreate(ns, pullRequestSender,
-                contentsWithImage, gitForkBranch, dockerfileGitHubUtil);
+                contentsWithImage, gitForkBranch, dockerfileGitHubUtil, rateLimiter);
         verify(all, times(1)).processErrorMessages(anyString(), anyString(), any());
         verify(all, times(1)).printSummary(anyList(), any());
```
cjneasbi, line 226: So you don't need `times(1)`.  `times(1)` is the default.

cjneasbi (changes requested):

dockerfile-image-update/pom.xml
```diff
@@ -139,6 +139,11 @@
             <version>3.24.0</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.bucket4j</groupId>
+            <artifactId>bucket4j-core</artifactId>
+            <version>8.0.1</version>
```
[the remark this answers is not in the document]
tarishij17: thanks for calling that out. the new version was released shortly after I started testing, so didn't realised that it's already out there

tarishij17:
> @tarishij17 Can you point to the unit test that validates that PR's are cut only at the rate configured by the rate limiter? It looks like the current unit tests were updated but no new ones were added.

I haven't added them yet. I wanted to have first review in place to incorporate any major changes before adding any unit tests.. will be adding in next commit.

jeetchoudhary:
- can we keep the limit configurable and have a default number a little higher? 
- does it make sense to stop the execution(or have an option) if a limit is reached for a given hour? with 30 PRs per hour we may not even come out of single execution if we keep on updating for the next hour

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -81,6 +82,19 @@ static ArgumentParser getArgumentParser() {
...
                 .setDefault(false)
                 .help("Only update image tag store. Skip creating PRs");
+        parser.addArgument("-u", "--" + Constants.USE_RATE_LIMITING)
+                .type(Boolean.class)
+                .setDefault(false)
+                .help("Use RateLimiting when sending PRs");
+        parser.addArgument("-r", "--" + Constants.RATE_LIMIT)
+                .type(Long.class)
+                .help("Max number of PRs to be sent per " + Constants.RATE_LIMIT_DURATION);
+        parser.addArgument("-rl", "--" + Constants.RATE_LIMIT_DURATION)
```
afalko, line 92: Is `-d` taken?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -81,6 +82,19 @@ static ArgumentParser getArgumentParser() {
                 .type(Boolean.class)
                 .setDefault(false)
                 .help("Only update image tag store. Skip creating PRs");
+        parser.addArgument("-u", "--" + Constants.USE_RATE_LIMITING)
```
afalko, line 85: Why do we need this flag? If rate limit is not set, then this would be false.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -81,6 +82,19 @@ static ArgumentParser getArgumentParser() {
...
+                .type(Boolean.class)
+                .setDefault(false)
+                .help("Use RateLimiting when sending PRs");
+        parser.addArgument("-r", "--" + Constants.RATE_LIMIT)
+                .type(Long.class)
+                .help("Max number of PRs to be sent per " + Constants.RATE_LIMIT_DURATION);
+        parser.addArgument("-rl", "--" + Constants.RATE_LIMIT_DURATION)
+                .type(Duration.class)
+                .help("Rate limit duration");
```
afalko, line 94: Let's consolidate this into the rate limit flag. You can represent durations via a set of characters such as `100m` (100 per minute), `500s` (500 per second), etc.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
afalko: Yes, the ide will generate post of this if you start your comment in the class with `/** `

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.github.bucket4j.Bandwidth;
+import io.github.bucket4j.Bucket;
+import io.github.bucket4j.Refill;
+import java.time.Duration;
+
+//based on token-bucket algorithm
+public class RateLimiter {
+
+    private long rateLimit;
```
afalko, line 11: This should be `final`, I don't think we want to ever change the rate limit, right?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,45 @@
...
+    private Duration rateLimitDuration;
```
afalko, line 12: Also needs to be `final`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,45 @@
...
+    private Duration tokenAddingRate;
```
afalko, line 13: This needs to be `final`

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,45 @@
...
+    private Bucket bucket;
```
afalko, line 14: Also make `final`

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,37 @@
...
+            @Override
+            public void run() {
+                try {
+                    RateLimiter rateLimiter = new RateLimiter(2,Duration.ofMinutes(2),Duration.ofMinutes(1));
+                    rateLimiter.consume();
+                    Thread t = Thread.currentThread();
+                    assertEquals(t.getState().toString(),"RUNNABLE");
+                    //thread will go in waiting state as the second token will be available only after 1 min
+                    rateLimiter.consume();
+                }catch(InterruptedException e){
```
afalko, line 23: Please configure your ide w/ google java code stype. You need spaces between the braces and `catch(...)`

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,37 @@
+package com.salesforce.dockerfileimageupdate.utils;
+
+import org.testng.annotations.Test;
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.assertEquals;
+import java.time.Duration;
+import java.lang.InterruptedException;
+
+public class RateLimiterTest extends Thread{
```
afalko, line 9: I don't think this needs to extend thread

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,37 @@
...
+            public void run() {
+                try {
+                    RateLimiter rateLimiter = new RateLimiter(2,Duration.ofMinutes(2),Duration.ofMinutes(1));
+                    rateLimiter.consume();
+                    Thread t = Thread.currentThread();
+                    assertEquals(t.getState().toString(),"RUNNABLE");
+                    //thread will go in waiting state as the second token will be available only after 1 min
+                    rateLimiter.consume();
+                }catch(InterruptedException e){
+                    System.out.println("exception thrown");
```
afalko, line 24: Why swallow this? I'd make this a callable that returns an `Optinal<Exception>` and have the test assert that no exception thrown

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,37 @@
...
+                }
+            }
+        };
+        t1.start();
+        Thread.sleep(30000);
```
afalko, line 29: This is going to make this test really long. Let's shorten this sleep expectation.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.assertEquals;
+import java.time.Duration;
+import java.lang.InterruptedException;
+
+public class RateLimiterTest extends Thread{
+    
+    @Test
+    public void testConsume() throws InterruptedException{
+        Thread t1 = new Thread(){
```
afalko, line 13: I'd do this with a Callable and a Future: https://www.baeldung.com/java-runnable-callable#2-with-callable

afalko: You can have callable return a POJO that has an optional exception and returns a state depending on what happen.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
tarishij17: we have kept this flag to provide a default rate limit. If only this flag value is true, then the default rate Limit will be set.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
tarishij17: Can you please point me to some example for this? As far as I understand, duration can represent the time part only and not the rate https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence- so 100m would mean 100minutes, etc. 
Also, the Refill methods where this will be used, expects the duration and rate as separate arguments https://javadoc.io/doc/com.github.vladimir-bukhtoyarov/bucket4j-core/4.10.0/io/github/bucket4j/Refill.html . If we try to club them during input, then we would have to parse them as well at later stage.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/subcommands/impl/All.java
```diff
@@ -138,4 +154,21 @@ protected GitForkBranch getGitForkBranch(String image, String tag, Namespace ns)
...
+            rateLimit = ns.get(Constants.RATE_LIMIT);
+        }
+        if(ns.get(Constants.RATE_LIMIT_DURATION)!=null){
+            rateLimitDuration = Duration.parse(ns.get(Constants.RATE_LIMIT_DURATION));
+        }
+        if(ns.get(Constants.TOKEN_ADDING_RATE)!=null){
+            tokenAddingRate = Duration.parse(ns.get(Constants.TOKEN_ADDING_RATE));
+        }
+        return new RateLimiter(rateLimit, rateLimitDuration, tokenAddingRate);
+    }
```
cjneasbi, line 173: This looks like it was copied in at least two places?  Can't we can create a single factory class for creating rate limiters and just call it in both places?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -358,9 +358,19 @@ public GitHubJsonStore getGitHubJsonStore(String store) {
 
     public void createPullReq(GHRepository origRepo,
```
cjneasbi, line 359: So we should have some javadocs on this method, especially since the rate limiter parameter is nullable.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.testng.annotations.Test;
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.assertEquals;
+import java.time.Duration;
+import java.lang.InterruptedException;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.Callable;
+import java.util.Optional;
+
+public class RateLimiterTest {
```
cjneasbi, line 12: I would recommend revising this test.  Tests based upon synchronization using wall clock time tend to flap.  They also tend to make your test run longer.  For instance, this test alone will add at ;east another 23 seconds to every build.  Instead of using defined wall clock intervals to synchronize threads for testing its better to provide a deterministic mock clock to the threads and modify the clocks values during the test.  Bucket4j seems to have to ability to inject a custom clock, see https://bucket4j.com/8.1.1/toc.html#customizing-time-measurement-specify-custom-time-measurement-strategy .  It also implements mock clock in it own tests, https://github.com/bucket4j/bucket4j/blob/7477830cff62cef125226468814cf13fa51b530c/bucket4j-core/src/test/java/io/github/bucket4j/mock/TimeMeterMock.java#L7 .  I recommend we go the route of mocking the clock instead of adding sleeps for this test.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
afalko: The default rate limit should be none - I still don't think this is necessary

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
afalko: We shouldn't pass through Refill's interface and provide a simpler, more constrained one

PR #717, opened 2022-12-27
DFIU to support rate limiting the PR creation in a given amount of time.

jeetchoudhary:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -84,6 +85,31 @@ static ArgumentParser getArgumentParser() {
         parser.addArgument("-x")
                 .help("comment snippet mentioned in line just before FROM instruction for ignoring a child image. " +
                         "Defaults to 'no-dfiu'");
+        parser.addArgument("-u", "--" + Constants.USE_RATE_LIMITING)
```
jeetchoudhary, line 88: @afalko  we can combine two of the args(rate limit and rate limit duration) into one say 500s or 500m (500 per min or seconds). But I choose to keep it separated mainly because of 
1. By default this feature is not enabled.
2. Even if it is enabled, the program defaults to a reasonable value.
3. Adding tokenizing the user input and converting it to the required format will have additional maintenance overhead and is prone to bugs over time. So I feel having two arguments will pan out better than combing them into one and tokenizing them ourselves.

Let me know if that sounds reasonable. if not I'll combine two of them into one.

afalko: I'd rather it be just one rate limit flag exposed to the user. Same feedback I gave in https://github.com/salesforce/dockerfile-image-update/pull/588 . 

Having multiple parameters that work together actually adds additional edge conditions that we have to handle. There is additional cognitive load on our users as well. These are all also top-level flags in the command so the more flags you add the other existing flags are diluted.

README.md
```diff
@@ -141,6 +141,11 @@ named arguments:
   -s {true,false}, --skipprcreation {true,false}
                          Only update image tag store. Skip creating PRs
   -x X                   comment snippet mentioned in line just before FROM instruction for ignoring a child image. Defaults to 'no-dfiu'
+  -u, --useratelimiting  {true,false}. Defaults to false, means no ratelimiting is imposed if this is not set to true
+                         Use RateLimiting for throttling the number of PRs DFIU will cut over a period of time. Throttling works based on token-bucket algorithm.
```
afalko, line 145: Users shouldn't be exposed to what algo is used for throttling. We can put in this in md docs in the repo, but not in the cmd-line interface

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
afalko: Just to clarify, I would prefer we combine all 4 into one param.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test(dataProvider = "rateLimitingData")
+    public void testingAgainstWallClock(int rateLimit, int durationLimit, int tokenAddingRate) throws Exception {
+        FutureTask futureTask = new FutureTask(new RateLimiterCallable
+                (new RateLimiterEvent(null, rateLimit, durationLimit, tokenAddingRate)));
+        Thread t1 = new Thread(futureTask);
+        long startTime = System.currentTimeMillis();
+        t1.start();
+        Thread.sleep(1000);
```
afalko, line 27: Is it possible to make this test take 100ms instead of 1sec?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+
+class RateLimiterCallable implements Callable {
+    RateLimiterEvent rateLimiterEvent;
+
+    RateLimiterCallable(RateLimiterEvent rateLimiterEvent) {
+        this.rateLimiterEvent = rateLimiterEvent;
+    }
+
+    @Test
```
afalko, line 70: Why do you need the @Test annotation here?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+                rateLimiter.consume();
+                rateLimiterEvent.incrementEventProcessed();
+                assertEquals(Thread.currentThread().getState().toString(), "RUNNABLE");
+                if (RateLimiterEvent.clock != null) {
+                    RateLimiterEvent.clock.addSeconds(rateLimiterEvent.tokenAddingRate);
+                }
+            }
+            return Optional.ofNullable(null);
+        } catch (InterruptedException e) {
+            System.out.println("exception thrown");
```
afalko, line 87: Why are we printing to STDOUT? This'll look out of context if you're debugging tests, especially if they run in parallel. In the worst case `log.error` or something like that - or delgate to the test caller to throw the exception via junit

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+            return Optional.ofNullable(new InterruptedException());
+        }
+    }
+}
+
+
+class RateLimiterEvent {
+    static CustomTimeMeter clock;
```
afalko, line 95: all these should be set to `final`

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+    long ratelimit;
+    long durationLimit;
+    long tokenAddingRate;
+    volatile int processedEventCount = 0;
```
afalko, line 99: Make this `AtomicInteger` instead of using volatile and also make `final`

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+    public void incrementEventProcessed() {
+        this.processedEventCount++;
+    }
+
+    public int getProcessedEventCount() {
+        return processedEventCount;
+    }
+}
+
+class CustomTimeMeter implements TimeMeter {
```
afalko, line 117: Is there a better way to describe this than `Custom`?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,139 @@
...
+    private volatile long currentTimeNanos;
```
afalko, line 118: Use AtomicLong and make `final` - remove volatile

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
afalko: Why is clock static?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
jeetchoudhary: let me work on it today and see what best to merge all of them.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
jeetchoudhary: Used MockTimeMeter.. does that looks okay?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
jeetchoudhary: the class was copied from the other pr. fixed it

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -357,10 +357,20 @@ public GitHubJsonStore getGitHubJsonStore(String store) {
...
     public void createPullReq(GHRepository origRepo,
-                              String branch, GHRepository forkRepo,
-                              PullRequestInfo pullRequestInfo) throws InterruptedException, IOException {
+                              String branch,
+                              GHRepository forkRepo,
+                              PullRequestInfo pullRequestInfo,
+                              RateLimiter rateLimiter) throws InterruptedException, IOException {
         // TODO: This may loop forever in the event of constant -1 pullRequestExitCodes...
         while (true) {
+            if(rateLimiter != null) {
```
cjneasbi, line 366: Suggestion.  Instead of checking for `null` instead assign a NullRateLimiter whose implementation of consume just returns immediately.
```
final RateLimiter localRateLimiter = Optional.ofNullable(rateLimiter).orElse(new NullRateLimiter())
log.info("Trying to consume a token before creating pull request..");
// Consume a token from the token bucket.
// If a token is not available this method will block until
// the refill adds one to the bucket.
rateLimiter.consume();
log.info("Token consumed, proceeding with PR creation..");
``` 
Then we don't need this branching logic.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * This method will create and return an object of type RateLimiter based on the
+     * variable set in the Namespace.
+     *
+     * @param ns Namespace {@see net.sourceforge.argparse4j.inf.Namespace}
+     * @return RateLimiter object if required variable is set in Namespace
+     *          null otherwise.
+     * @see net.sourceforge.argparse4j.inf.Namespace Namespace
+     */
+    public RateLimiter getRateLimiter(Namespace ns) {
```
cjneasbi, line 95: This can be static correct?  Looks like a factory method.  Can we return an Optional<RateLimiter> instead of returning null?

README.md
```diff
@@ -213,6 +216,34 @@ Example:
...
 
+### PR throttling
+
+In case you want to throttle the number of PRs cut by DFIU over a period of time,
+set --rate-limit-pr-creations with appropriate value. More example below
+
+##### Default case:
+
+```
+example usage: dockerfile-image-update --useratelimiting 500 all image-tag-store-repo-falcon
```
afalko, line 227: update to new single flag

README.md
```diff
@@ -141,7 +141,10 @@ named arguments:
   -s {true,false}, --skipprcreation {true,false}
                          Only update image tag store. Skip creating PRs
   -x X                   comment snippet mentioned in line just before FROM instruction for ignoring a child image. Defaults to 'no-dfiu'
-
+  -r, --rate-limit-pr-creations 
+                         Enable rateLimiting for throttling the number of PRs DFIU will cut over a period of time.
+                         Defaults is not set, means no ratelimiting is imposed if this is not set to true
```
afalko, line 146: Isn't the default 500-per-h or 500?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+ * time, the number of tokens in the bucket won't exceed the 'rateLimit', limiting the rate at
+ * which PRs would be raised. To keep the refill rate uniform, one token will be added to the
+ * bucket every 'tokenAddingRate'. If no tokens are left in the bucket, then PR won't be raised
+ * and program will halt until next token is available.
+ * <pre>
+ */
+public class RateLimiter {
+
+    private static final Logger log = LoggerFactory.getLogger(RateLimiter.class);
+    private final long rateLimit;
```
afalko, line 34: rateLimit is duplicate with `rateLimitDuration`....can you make a single 

`Duration rateLimit;`

afalko: rateLimitDuration is already initialized with the rate

afalko: specifically here: https://github.com/salesforce/dockerfile-image-update/pull/717/files#diff-b17e0a9f060287f7ce7cd459e67a05449d5980ee6faddf7c31e8ac1c6848e6a0R45

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+                case "h":
+                case "H":
+                    rLimitDuration = Duration.ofHours(duration);
+                    tokAddingRate = Duration.ofHours(rLimit / duration);
+                    break;
+                default:
+                    //should not reach here are regex will enforce char, keeping it for any unexpected use case.
+                    throw new UnknownFormatConversionException("Unexpected format encountered, can't tokenize input");
+            }
+            return new RateLimitEvent(rLimit, rLimitDuration, tokAddingRate);
```
afalko, line 221: same note here, I think `rLimit` carries duplicate data with `rLimitDuration` and should be consolidated into one

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+                    tokAddingRate = Duration.ofMinutes(rLimit / duration);
+                    break;
+                case "h":
+                case "H":
+                    rLimitDuration = Duration.ofHours(duration);
+                    tokAddingRate = Duration.ofHours(rLimit / duration);
+                    break;
+                default:
+                    //should not reach here are regex will enforce char, keeping it for any unexpected use case.
+                    throw new UnknownFormatConversionException("Unexpected format encountered, can't tokenize input");
```
afalko, line 219:
```suggestion
                    throw new UnknownFormatConversionException("Unexpected format or unit encountered, valid input is <integer>-per-<time-unit> where time-init is one of 's', 'm', or 'h'. Example: 500-per-h");
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+            Duration rLimitDuration;
+            Duration tokAddingRate;
+            String literalConstant;
+            String rateDuration;
+            String rateDurationUnitChar;
+
+            Matcher matcher = eventPattern.matcher(input);
+
+            if (!matcher.matches()) {
+                throw new UnknownFormatConversionException("input argument is not in valid format and can't be tokenize");
```
afalko, line 176: Same suggestion: https://github.com/salesforce/dockerfile-image-update/pull/717/files#r1059085290

afalko: You also have duplicate text - would be good to have the same error msg returned

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,172 @@
...
+import static org.testng.Assert.*;
+
+public class RateLimiterTest {
+
+    @DataProvider(name = "rateLimitingData")
+    public static Object[][] rateLimitingData() {
+        return new Object[][]{{2, 10, 5}, {4, 16, 4}};
+    }
+
+    @DataProvider(name = "getRateLimitingData")
```
afalko, line 23: confusing data provider name... can you also move this down to where the test is?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+                 * 1 group is numeric rate limit value and will always
+                 *                          be present if exp matched
+                 * 2 is literal constant '-per-' and is optional
+                 * 3 is numeric rate duration and is optional
+                 * 4 is a char(s/m/h) and is optional
+                 */
+                rLimit = Long.parseLong(matcher.group(1)); // will always have a numeric value that fits in Long
+                literalConstant = matcher.group(2); // can be null or empty or '-per-'
+                rateDuration = matcher.group(3);// can be null or empty or numeric value
+                rateDurationUnitChar = matcher.group(4); //can be empty or char s,m,h
```
cjneasbi, line 190: You probably want to test that the expected number of groups exist before calling group.  Otherwise this will throw an IndexOutofBoundsException , https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#group(int)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            if (literalConstant == null || literalConstant.isEmpty()) {
```
cjneasbi, line 193: Apache utils a nice method for this, https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isEmpty-java.lang.CharSequence-

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -0,0 +1,224 @@
...
+                    break;
+                case "h":
+                case "H":
+                    rLimitDuration = Duration.ofHours(duration);
+                    tokAddingRate = Duration.ofHours(rLimit / duration);
+                    break;
+                default:
+                    //should not reach here are regex will enforce char, keeping it for any unexpected use case.
+                    throw new UnknownFormatConversionException("Unexpected format encountered, can't tokenize input");
+            }
```
cjneasbi, line 220: Is there a reason why we are building our own duration parser instead of using Java's duration parsing APIs.  See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalTime.html#parse(java.lang.CharSequence) and https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html#parse(java.lang.CharSequence) .  See https://stackoverflow.com/questions/8257641/java-how-to-convert-a-string-hhmmss-to-a-duration for details.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ChildTest.java
```diff
@@ -69,19 +76,21 @@ public void checkPullRequestMadeForS3ImageStore() throws Exception {
                 GIT_REPO, "test",
                 IMG, "test",
                 FORCE_TAG, "test",
-                STORE, "s3://test");
+                STORE, "s3://test"
+               // USE_RATE_LIMITING, false);
```
cjneasbi, line 80: Is this commented code?  We might want to remove it.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
```diff
@@ -0,0 +1,172 @@
...
+        Thread t1 = new Thread(futureTask);
+        t1.start();
+        Thread.sleep(100);
+        // the token should be consumed within first 3s leaving the thread in blocked state as it'd be waiting for
+        // next token to be available
+        assertEquals(t1.getState().toString(), Thread.State.TIMED_WAITING.toString());
+        t1.join();
+        assertEquals(t1.getState().toString(), Thread.State.TERMINATED.toString());
+        assertEquals(futureTask.get(), Optional.ofNullable(null));
+    }
```
cjneasbi, line 54: Does this cover any use case that `testingAgainstCustomClock` doesn't?

jeetchoudhary: Testing against a real system clock. that depends on the current time in Millis instead of the user-provided clock. Make sense to still have it while all other tests depends on the custom clock

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
jeetchoudhary: we are asking customers to provide a non-ISO-8601 duration format that is more intuitive and contains multiple values in a string

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
jeetchoudhary: I encountered that when the matcher wouldn't match, so wrapped this code around the condition that we execute this only if the pattern matches. 
But I can still handle the exception anyway. let me know your thoughts

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/RateLimiterTest.java
jeetchoudhary: changed the name.. however, still didn't bring the data and method together. IDE default formatting takes it back to the top of the class.  That'll happen with most of the devs...

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
cjneasbi: Why not just ask for the standard format?  We'd write less code and handle more cases.

Here is the translation of the test cases to use the ISO-8601 standard duration instead of our custom format.  I don't think the inputs look all that different.  But the parsing code shrinks to basically `Duration.parse(input.split("-")[1])`
```
500-per-s" -> 500-PT1S
"500-per-60s" -> 500-PT60S
"500-per-1m" -> 500-PT1M
"500-per-1h" -> 500-PT1H
"1234-per-2h - 1234-PT2H
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
cjneasbi: So I did a quick test with a regex tester.  See https://www.regexplanet.com/share/index.html?share=yyyypr0qx0r .  Wouldn't an input like `1` cause matcher.matches() to be `true` but having less than the desired number of groups?  Maybe we can add a unit test for this.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
afalko: I like what @cjneasbi is recommending here - it is much cleaner and standardized

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
jeetchoudhary: The number of groups is determined by the regex, not by the matcher. The matcher fills in the data if it founds the match. 
for example 
    Pattern.compile("(^[1-9]\\d{0,18})(-per-)?([1-9]\\d{0,18})*([smh]?)").matcher("rando");
        System.out.println(m.matches());
        System.out.println(m.groupCount());
will print false and 4 despite the pattern doesn't match.  Same for value one. it will match and matcher group 1 will have the required value.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
jeetchoudhary: ahh.. thought of that while implementing and decided to go with the exact format suggested in the previous code review comments. Let me make the change and notify you guys as soon as I get time.
thanks

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
cjneasbi: Got it.  So in the case of the value "1" group 1 will return 1 but group 2 will return null and so forth.

justinharringa:
Is there a particular reason to do this? IIRC the GitHub library itself provides for adjustments and GitHub itself provides their rates in API calls. 
Mostly curious...

afalko:
> Is there a particular reason to do this? IIRC the GitHub library itself provides for adjustments and GitHub itself provides their rates in API calls. Mostly curious...

We're doing it to allow rate limiting if other downstream parts of our infrastructure are under pressure, e.g. your docker registry. The Github library only protects us from running over Github limits.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
jeetchoudhary: Two reasons why Optional wasn't used here. 
approach 1:  Optional with object value or null, in this case, we'll have to pass around the optional as a method argument because of how it's being used in the code, if we choose not to do that we will end up having similar logic. 
approach 2:  Use a custom/Null rate limiter that has infinite tokens to consume. I think that's an additional code invocation for something that we don't even need if it's disabled.
I would preferred to use it if we were not passing the return value as arguments to other methods.

jeetchoudhary: open to thoughts

jeetchoudhary:
@afalko  @cjneasbi do take another look and let me know if things look okay. the comments should be addressed in the. last commit. Thanks

README.md
afalko, line 146:
```suggestion
                         Default is not set, this means no ratelimiting is imposed.
```

README.md
```diff
@@ -141,7 +141,10 @@ named arguments:
   -s {true,false}, --skipprcreation {true,false}
                          Only update image tag store. Skip creating PRs
   -x X                   comment snippet mentioned in line just before FROM instruction for ignoring a child image. Defaults to 'no-dfiu'
-
+  -r, --rate-limit-pr-creations 
+                         Enable rateLimiting for throttling the number of PRs DFIU will cut over a period of time.
```
afalko, line 145:
```suggestion
                         Enable rateLimiting for throttling the number of PRs DFIU will cut over a period of time. The argument value should be in format "<positive_integer>-<ISO-8601_formatted_time>".
For example "--rate-limit-pr-creations 60-PT1H" to create 60 PRs per hour.
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
afalko: We can add a TODO: to clean this up later - I don't think it is too much of a big deal as is - keeping the current change smaller is kind of nice too

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
```diff
@@ -122,7 +125,7 @@ public void testParentCommandSuccessfulForS3ImageStore() throws Exception {
                 "image", Constants.TAG,
                 "tag", Constants.STORE,
                 "s3://store", Constants.SKIP_PR_CREATION,
-                false);
+                false, RATE_LIMIT_PR_CREATION, "500-per-60s");
```
afalko, line 128: This looks like old format, no?

afalko (changes requested):
Small suggestions remaining - this is pretty much ready to go

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/subcommands/impl/ParentTest.java
jeetchoudhary: the correct values weren't required for the test. removed it.

afalko (approved):

PR #776, opened 2023-01-31
Fix special char usage in cmd line args for Rate Limiting the PRs

tarishij17:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RateLimiter.java
```diff
@@ -75,7 +75,7 @@ public RateLimiter() {
      * @see net.sourceforge.argparse4j.inf.Namespace Namespace
      */
     public static RateLimiter getInstance(Namespace ns) {
-        String rateLimitPRCreation = ns.get(Constants.RATE_LIMIT_PR_CREATION);
+        String rateLimitPRCreation = ns.get((Constants.RATE_LIMIT_PR_CREATION).replace("-","_"));
```
jeetchoudhary, line 78: I think the better thing is to use the underscore in the constant string and add a unit test to validate that no char is used in ArgumentParser that will be replaced automatically. 
Also, we need to test the changes once again.

afalko: What is the error that we are hitting if we don't make this change? Can we change the documentation is error out if invalid chars are passed?

jeetchoudhary: '-' is replaced by '_' by argparse4j ArgumentParser while parsing arguments. so "random-arg val" will be set to random_arg=val in the namespace.
What we are purposing here is 
1. change the argument using - to use _
2. add a unit test to throw the error if someone uses it unintentionally.

afalko: I see, so users won't see any difference, it is just how we parse internally?

afalko (approved):

jeetchoudhary (approved):
LGTM

PR #874, opened 2023-04-03
Skip sending PRs if no files in the repository is modified

avimanyum:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -277,17 +277,20 @@ public GHBlob tryRetrievingBlob(GHRepository repo, String path, String branch)
...
+    public boolean modifyOnGithub(GHContent content,
                                String branch, String img, String tag,
                                String customMessage, String ignoreImageString) throws IOException {
+        boolean modified;
         try (InputStream stream = content.read();
              InputStreamReader streamR = new InputStreamReader(stream);
              BufferedReader reader = new BufferedReader(streamR)) {
-            findImagesAndFix(content, branch, img, tag, customMessage, reader, ignoreImageString);
+             modified = findImagesAndFix(content, branch, img, tag, customMessage, reader,
+                    ignoreImageString);
```
cjneasbi, line 288: Why do you need the `modified` variable?  The return can be within the try-with-resources correct?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -542,9 +546,12 @@ public void changeDockerfiles(Namespace ns,
...
                 log.info("No Dockerfile found at path: '{}'", pathToDockerfile);
             } else {
-                modifyOnGithub(content, gitForkBranch.getBranchName(), gitForkBranch.getImageName(), gitForkBranch.getImageTag(),
-                        ns.get(Constants.GIT_ADDITIONAL_COMMIT_MESSAGE), ns.get(Constants.IGNORE_IMAGE_STRING));
-                isContentModified = true;
+                if (modifyOnGithub(content, gitForkBranch.getBranchName(),
+                        gitForkBranch.getImageName(), gitForkBranch.getImageTag(),
+                        ns.get(Constants.GIT_ADDITIONAL_COMMIT_MESSAGE),
+                        ns.get(Constants.IGNORE_IMAGE_STRING))) {
+                    isContentModified = true;
```
cjneasbi, line 553:
```
isContentModified = modifyOnGithub(content, gitForkBranch.getBranchName(),
                        gitForkBranch.getImageName(), gitForkBranch.getImageTag(),
                        ns.get(Constants.GIT_ADDITIONAL_COMMIT_MESSAGE),
                        ns.get(Constants.IGNORE_IMAGE_STRING)));
```
You don't need the conditional here.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java
```diff
@@ -880,6 +888,148 @@ public void testOnePullRequestForMultipleDockerfilesInSameRepo() throws Exceptio
...
+
+        // Dockerfile retrieved from the repo
+        verify(dockerfileGitHubUtil).tryRetrievingContent(eq(forkedRepo),
+                eq("df11"), eq("image-tag"));
+
+        // Trying to modify Dockerfile
+        verify(dockerfileGitHubUtil).modifyOnGithub(any(), eq("image-tag"), eq("image"), eq("tag"), any(), any());
+
+        // PR creation block not executed
+        verify(dockerfileGitHubUtil, times(0)).createPullReq(eq(parentRepo),
```
cjneasbi, line 950: Prefer `never()` to `times(0)`, https://stackoverflow.com/questions/12862659/how-to-verify-that-a-specific-method-was-not-called-using-mockito

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
avimanyum: This will not work - the modifyOnGithub will be called for all GHContent. In this case, the `isContentModified` will have the value returned for the last content on that loop. However, the idea is that the flag should be set to `true` when any content is modified.

cjneasbi: Got it.  The loop wasn't displayed.  You can aggregate with logical or.  See https://stackoverflow.com/questions/2486472/shortcut-or-assignment-operator-in-java .  According to the documentation `|=` should evaluate both operands so you don't have to worry about shortcutting.
```
isContentModified |= modifyOnGithub(content, gitForkBranch.getBranchName(),
                        gitForkBranch.getImageName(), gitForkBranch.getImageTag(),
                        ns.get(Constants.GIT_ADDITIONAL_COMMIT_MESSAGE),
                        ns.get(Constants.IGNORE_IMAGE_STRING)));
```

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -567,6 +574,8 @@ public void changeDockerfiles(Namespace ns,
                     forkedRepo,
                     pullRequestInfo,
                     rateLimiter);
+        } else {
+            log.info("No files changed in repo {}. Skipping PR creation attempt.", parentName);
```
afalko, line 578: This contains the org name as well rightr?

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtilTest.java
```diff
@@ -880,6 +888,148 @@ public void testOnePullRequestForMultipleDockerfilesInSameRepo() throws Exceptio
...
+
+        GHRepository forkedRepo = mock(GHRepository.class);
+        when(forkedRepo.isFork()).thenReturn(true);
+        when(forkedRepo.getFullName()).thenReturn("forkedrepo");
+
+        GHRepository parentRepo = mock(GHRepository.class);
+        when(parentRepo.getFullName()).thenReturn("repo");
+        when(forkedRepo.getParent()).thenReturn(parentRepo);
+
+        //GHRepository parent = mock(GHRepository.class);
```
afalko, line 910: Remove this please

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
avimanyum: yes

afalko (approved):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/DockerfileGitHubUtil.java
```diff
@@ -280,14 +280,21 @@ public GHBlob tryRetrievingBlob(GHRepository repo, String path, String branch)
     public void modifyOnGithub(GHContent content,
                                String branch, String img, String tag,
                                String customMessage, String ignoreImageString) throws IOException {
+        modifyContentOnGithub(content, branch, img, tag, customMessage, ignoreImageString);
+    }
+
+    protected boolean modifyContentOnGithub(GHContent content,
```
cjneasbi, line 286: Utility classes are typically labeled as `final` and have private constructors to prevent extension.  We didn't do that in the case of this class.  As such I would make these methods private or package private if possible to prevent customers taking a dependency on them.

cjneasbi (approved):

afalko (approved):

PR #983, opened 2023-06-13
feat: add "dependencies" labels to DFIU-created PRs

jasonschroeder-sfdc:
This adds "dependencies" label to DFIU-created PRs.

### DIFF
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
@@ -105,6 +105,7 @@ public class GitHubUtil {
         try {
             GHPullRequest pullRequest = origRepo.createPullRequest(title, forkRepo.getOwnerName() + ":" + branch,
                     origRepo.getDefaultBranch(), body);
+            pullRequest.setLabels("dependencies");
 //            origRepo.createPullRequest("Update base image in Dockerfile", forkRepo.getOwnerName() + ":" + branch,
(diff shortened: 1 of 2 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GitHubUtil.java
```diff
@@ -105,6 +105,7 @@ public int createPullReq(GHRepository origRepo, String branch,
         try {
             GHPullRequest pullRequest = origRepo.createPullRequest(title, forkRepo.getOwnerName() + ":" + branch,
                     origRepo.getDefaultBranch(), body);
+            pullRequest.setLabels("dependencies");
```
jeetchoudhary, line 108: can we run it against an org/repo that doesn't have **dependencies** label already created? Not sure If the will auto-create a new one  If it doesn't exist. In the earlier version of GHE, I believe we had to create a label explicitly first and then apply it. 
LGTM otherwise

jasonschroeder-sfdc: As I have seen from other tools, the label would be auto-created.
I read https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#about-labels and didn't see anything about the label needing to exist first.

jeetchoudhary: wanted to be sure about it. vaguely remember this was needed a couple of years back. 
Tested it on test-temp/test/pull/1 on soma. works fine. we should be good here

jeetchoudhary (approved):

afalko (approved):

PR #1128, opened 2024-02-07
Skip DFIU PRs for repos onboarded to Renovate Enterprise

avimanyum:
Renovate enterprise is another solution that can be used to receive remediation PRs. This change disables sending DFIU PRs for those repositories that have been onboarded to Renovate Enterprise.

### DIFF
--- a/dockerfile-image-update/pom.xml
+++ b/dockerfile-image-update/pom.xml
@@ -72,7 +72,7 @@
         <dependency>
             <groupId>org.json</groupId>
             <artifactId>json</artifactId>
-            <version>20200518</version>
+            <version>20240205</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/dockerfile-image-update/pom.xml
+++ b/dockerfile-image-update/pom.xml
@@ -70,6 +70,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20200518</version>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
             <version>${slf4j.version}</version>
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
@@ -65,7 +65,7 @@ public class CommandLine {
                 ArgumentParsers.newFor("dockerfile-image-update").addHelp(true).build()
                 .description("Image Updates through Pull Request Automator");
 
-        parser.addArgument("-re", "--" + CHECK_FOR_RENOVATE)
+        parser.addArgument("-R", "--" + CHECK_FOR_RENOVATE)
                 .type(Boolean.class)
                 .setDefault(false) //To prevent null from being returned by the argument
                 .help("Check if Renovate app is being used to receive remediation PRs.");
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
@@ -65,6 +65,10 @@ public class CommandLine {
                 ArgumentParsers.newFor("dockerfile-image-update").addHelp(true).build()
                 .description("Image Updates through Pull Request Automator");
 
+        parser.addArgument("-re", "--" + CHECK_FOR_RENOVATE)
+                .type(Boolean.class)
+                .setDefault(false) //To prevent null from being returned by the argument
+                .help("Check if Renovate app is being used to receive remediation PRs.");
         parser.addArgument("-l", "--" + GIT_API_SEARCH_LIMIT)
                 .type(Integer.class)
                 .setDefault(1000)
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
@@ -54,5 +54,5 @@ public class Constants {
 
     public static final String CHECK_FOR_RENOVATE = "checkforrenovate";
 
-    public static final String RENOVATE_CONFIG_FILENAME = "renovate.json";
+    public static final String RENOVATE_CONFIG_FILEPATH = "renovate.json";
 }
--- a/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
+++ b/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
@@ -10,6 +10,7 @@ package com.salesforce.dockerfileimageupdate.utils;
 
 
 import java.time.Duration;
+import java.util.*;
 
 /**
  * @author minho-park
@@ -52,5 +53,6 @@ public class Constants {
     public static final String FILENAME_DOCKERFILE = "dockerfile";
     public static final String FILENAME_DOCKER_COMPOSE = "docker-compose";
     public static final String CHECK_FOR_RENOVATE = "checkforrenovate";
-    public static final String RENOVATE_CONFIG_FILEPATH = "renovate.json";
+    //The Renovate configuration file can be in any one of the following locations. Refer to https://docs.renovatebot.com/configuration-options/
+    public static final List<String> RENOVATE_CONFIG_FILEPATHS = Arrays.asList("renovate.json", "renovate.json5", ".github/renovate.json", ".github/renovate.json5", ".gitlab/renovate.json", ".gitlab/renovate.json5", ".renovaterc", ".renovaterc.json", ".renovaterc.json5");
 }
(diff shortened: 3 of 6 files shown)


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
```diff
@@ -51,4 +51,6 @@ private Constants() {
     public static final Duration DEFAULT_TOKEN_ADDING_RATE = Duration.ofMinutes(DEFAULT_CONSUMING_TOKEN_RATE);
     public static final String FILENAME_DOCKERFILE = "dockerfile";
     public static final String FILENAME_DOCKER_COMPOSE = "docker-compose";
+    public static final String CHECK_FOR_RENOVATE = "checkforrenovate";
+    public static final String RENOVATE_CONFIG_FILEPATH = "renovate.json";
```
cjneasbi, line 55: While we will using `renovate.json` internally there are actually 9 different potential renovate configuration files,  see https://docs.renovatebot.com/configuration-options/ .  We might want to check for all of these paths.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,31 @@ public void prepareToCreate(final Namespace ns,
...
+     * (2) Ensuring that the file does not have the key "enabled" set to "false"
+     * @param filePath the name of the file that we are searching for in the repo
+     * @param fork A GitHubContentToProcess object that contains the fork repository that is under process
+     * @return true if the file is found in the path specified and is not disabled, false otherwise
+     */
+    private boolean isRenovateEnabled(String filePath, GitHubContentToProcess fork) throws IOException {
+        try {
+            GHContent fileContent = fork.getParent().getFileContent(filePath);
+            InputStream is = fileContent.read();
+            BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(is));
```
cjneasbi, line 65: I think these should be closed after use.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,31 @@ public void prepareToCreate(final Namespace ns,
...
+            JSONTokener tokener = new JSONTokener(bufferedReader);
+            JSONObject json = new JSONObject(tokener);
+            //If the renovate.json file has the key 'enabled' set to false, it indicates that while the repo has been onboarded to renovate, it has been disabled for some reason
+            if (json.has("enabled") && json.getBoolean("enabled") == false) {
+                return false;
+            }
```
cjneasbi, line 71: I think you can replace this if statement with

```java
return json.optBoolean("enabled", true)
```
If the file exists and enabled is not set it is assumed to be true, otherwise return the value of the enabled field.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,31 @@ public void prepareToCreate(final Namespace ns,
...
+        } catch (FileNotFoundException e) {
+            log.debug("The file with name %s not found in the repository.", filePath);
+            return false;
+        }
+        return true;
```
cjneasbi, line 76: I think you can remove the `return` inside of the catch and change the last `return` to `return false`.  If the file isn't found should be only way the last return is executed so we should assume that renovate is not enabled.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -65,6 +65,10 @@ static ArgumentParser getArgumentParser() {
                 ArgumentParsers.newFor("dockerfile-image-update").addHelp(true).build()
                 .description("Image Updates through Pull Request Automator");
 
+        parser.addArgument("-re", "--" + CHECK_FOR_RENOVATE)
```
cjneasbi, line 68: What about `-R` instead for the short opt?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,32 @@ public void prepareToCreate(final Namespace ns,
...
+    private boolean isRenovateEnabled(List<String> filePaths, GitHubContentToProcess fork) throws IOException {
+        for (String filePath : filePaths) {
+            try {
+                GHContent fileContent = fork.getParent().getFileContent(filePath);
+                InputStream is = fileContent.read();
+                BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(is));
+                JSONTokener tokener = new JSONTokener(bufferedReader);
+                JSONObject json = new JSONObject(tokener);
+                is.close();
+                bufferedReader.close();
```
cjneasbi, line 70: You might want to do this in a try-with-resources statement, see https://stackoverflow.com/a/56151320 .  If `is.close` throws an exception then the close on buffered reader won't happen.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,34 @@ public void prepareToCreate(final Namespace ns,
...
+     * @param fork A GitHubContentToProcess object that contains the fork repository that is under process
+     * @return true if the file is found in the path specified and is not disabled, false otherwise
+     */
+    private boolean isRenovateEnabled(List<String> filePaths, GitHubContentToProcess fork) throws IOException {
+        for (String filePath : filePaths) {
+            try {
+                GHContent fileContent = fork.getParent().getFileContent(filePath);
+                JSONObject json;
+                try (InputStream is = fileContent.read();
+                     BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(is))) {
```
cjneasbi, line 67: I think the input stream is closed when the buffered reader is closed so I think you can inline `fileContent.read()` in the `InputStreamReader` constructor.

cjneasbi (dismissed):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -41,5 +49,33 @@ public void prepareToCreate(final Namespace ns,
...
+    protected boolean isRenovateEnabled(List<String> filePaths, GitHubContentToProcess fork) throws IOException {
+        for (String filePath : filePaths) {
+            try {
+                GHContent fileContent = fork.getParent().getFileContent(filePath);
+                JSONObject json;
+                try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(fileContent.read()))) {
+                    JSONTokener tokener = new JSONTokener(bufferedReader);
+                    json = new JSONObject(tokener);
+                    //If the file has the key 'enabled' set to false, it indicates that while the repo has been onboarded to renovate, it has been disabled for some reason
+                    return json.optBoolean("enabled", true);
```
jeetchoudhary, line 70: So the users need to delete this key altogether to get DFIU PRs back? true/false both mean we skip DFIU PRs?

avimanyum: Here are the scenarios
- Renovate config file not found - Send DFIU PR
- Renovate config file found and file does not have key 'enabled' - Skip DFIU PR
- Renovate config file found and file has the key 'enabled' - return the boolean value for that key. If set to "true", skip DFIU PR, if set to "false", send DFIU PR

This key is usually used when a repo onboards to Renovate but then decides to off board by adding the key `enabled`: false.

dockerfile-image-update/pom.xml
```diff
@@ -69,6 +69,11 @@
 
 
     <dependencies>
+        <dependency>
+            <groupId>org.json</groupId>
```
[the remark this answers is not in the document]
avimanyum: Yeah its just a cleaner and easier way to convert to Json object

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
jeetchoudhary: "Renovate config file found and file does not have key enabled - Skip DFIU PR" Should we do this? the config file doesn't have this key, doesn't that mean it's enabled? the default is set to true https://docs.renovatebot.com/configuration-options/#enabled

jeetchoudhary: and i don't think the method works like that at the moment. https://github.com/stleary/JSON-java/blob/master/src/main/java/org/json/JSONObject.java#L1143
The method will return true if the file doesn't have the key enabled and we'll skip the PR

jeetchoudhary: or am I overlooking something?

avimanyum: > "Renovate config file found and file does not have key enabled - Skip DFIU PR" Should we do this? the config file doesn't have this key, doesn't that mean it's enabled? the default is set to true https://docs.renovatebot.com/configuration-options/#enabled

Yes, it does mean it is enabled. So we will skip DFIU because we expect to receive PRs from Renovate

jeetchoudhary (approved):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
jeetchoudhary: I see. This comment confused me a little, I thought we wanted to skip sending the PR in this case. 
//If the file has the key 'enabled' set to false, it indicates that while the repo has been onboarded to renovate, it has been disabled for some reason.

PR #1130, opened 2024-02-13
Handle JSONException while reading Renovate configuration file

avimanyum:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -67,6 +68,8 @@ protected boolean isRenovateEnabled(List<String> filePaths, GitHubContentToProce
                 log.debug("The file with name {} not found in the repository. Exception: {}", filePath, e.getMessage());
             } catch (IOException e) {
                 log.debug("Exception while trying to close a resource. Exception: {}", e.getMessage());
+            } catch (JSONException e) {
+                log.debug("Exception while trying to read the renovate configuration file. Exception: {}", e.getMessage());
```
cjneasbi, line 72: Its helpful to state what the expected return value is in the error message for this situation.  For example we can add the phase `returning false` to the message.

cjneasbi (dismissed):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
avimanyum: @cjneasbi Updated

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
```diff
@@ -64,9 +65,11 @@ protected boolean isRenovateEnabled(List<String> filePaths, GitHubContentToProce
                 //If the file has the key 'enabled' set to false, it indicates that while the repo has been onboarded to renovate, it has been disabled for some reason
                 return readJsonFromContent(fork.getParent().getFileContent(filePath)).optBoolean("enabled", true);
             } catch (FileNotFoundException e) {
-                log.debug("The file with name {} not found in the repository. Exception: {}", filePath, e.getMessage());
+                log.debug("The file with name {} not found in the repository.Returning false. Exception: {}", filePath, e.getMessage());
```
jeetchoudhary, line 68: FileNotFoundException error should be in the debug log because we are expecting this for most of the repos. But do we want to do that for IO and JSON exceptions as well? We should log it in warn mode if they are expected to be raised in corner cases

jeetchoudhary (approved):

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/PullRequests.java
avimanyum: @jeetchoudhary Updated

PR #1133, opened 2024-08-08
[DRAFT] Update DFIU to verify whether renovate is installed

AvvariSaiBharadwaj:


dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
```diff
@@ -42,6 +42,8 @@ private Constants() {
     public static final String IGNORE_IMAGE_STRING = "x";
     public static final String FILE_NAMES_TO_SEARCH = "filenamestosearch";
     public static final String RATE_LIMIT_PR_CREATION = "rate_limit_pr_creations";
+    public static final String RENOVATE_GITHUB_APP_ID = "appId";
```
afalko: This should be more generic than renovate. The logic is to skip DFIU if a particular github app is installed

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.security.KeyFactory;
+import java.security.Security;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.time.Instant;
+import java.util.Date;
+
+public class RenovateUtil {
```
afalko:
```suggestion
public class GithubAppCheck {
```

Util is not a good descriptive name

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
```diff
@@ -0,0 +1,103 @@
...
+            gitHub = new GitHubBuilder()
+                    .withEndpoint(CommandLine.gitApiUrl(ns))
+                    .withJwtToken(jwt)
+                    .build();
+        } catch (IOException exception) {
+            log.warn("Could not initialise github due to exception: {}", exception.getMessage());
+        }
+    }
+
+    protected boolean isRenovateEnabledOnRepository(String fullRepoName){
```
afalko: let's make this more generic - to check for any app existance

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
```diff
@@ -0,0 +1,103 @@
...
+import java.security.Security;
+import java.security.interfaces.RSAPrivateKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.time.Instant;
+import java.util.Date;
+
+public class RenovateUtil {
+    private static final Logger log = LoggerFactory.getLogger(RenovateUtil.class);
+
+    private static String appId;
```
afalko: these all need to be made `final` static is not a good practice

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/Constants.java
AvvariSaiBharadwaj: 100%. WIll refactor the entire code to make it more generic.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
```diff
@@ -0,0 +1,103 @@
...
+    private static String privateKeyPath;
+    private static String jwt;
+    private static Instant jwtExpiry;
+    private static GitHub gitHub;
+
+    public RenovateUtil(final Namespace ns){
+        this.appId = ns.get(Constants.RENOVATE_GITHUB_APP_ID);
+        this.privateKeyPath = ns.get(Constants.RENOVATE_GITHUB_APP_KEY);
```
ghost: Why do we have to do `ns.get()` in this case instead of calling directly from Constants?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
```diff
@@ -0,0 +1,103 @@
...
+                    .build();
+        } catch (IOException exception) {
+            log.warn("Could not initialise github due to exception: {}", exception.getMessage());
+        }
+    }
+
+    protected boolean isRenovateEnabledOnRepository(String fullRepoName){
+        refreshJwtIfNeeded(appId, privateKeyPath);
+        try {
+            gitHub.getApp().getInstallationByRepository(fullRepoName.split("/")[0], fullRepoName.split("/")[1]);
```
ghost: From the owner's open source, looks like it is using GHApp instead of GithubBuilder for this logic: https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHApp.java

I don't see such getApp() function be mentioned in GithubBuilder: https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubBuilder.java

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: Done.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: Done.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: Done.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: This is to call the CLI input to the Java code. The constant is the CLI identifier for the input.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: I believe this is the one: https://github.com/hub4j/github-api/blob/af5bde714c3f771f6793392343ff0187bd8180df/src/main/java/org/kohsuke/github/GitHub.java#L1184. We are calling the method from the object type Github.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
```diff
@@ -0,0 +1,103 @@
...
+            this.gitHub = null;
+        }
+    }
+
+    protected boolean isGithubAppEnabledOnRepository(String fullRepoName){
+        refreshJwtIfNeeded(appId, privateKeyPath);
+        try {
+            gitHub.getApp().getInstallationByRepository(fullRepoName.split("/")[0], fullRepoName.split("/")[1]);
+            return true;
+        } catch (HttpException exception) {
```
jeetchoudhary, line 70: If we're defaulting to false, there's a risk of false negatives.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
```diff
@@ -0,0 +1,103 @@
...
+            return false;
+        }
+    }
+
+    private void refreshJwtIfNeeded(String appId, String privateKeyPath){
+        if (jwt == null || jwtExpiry.isBefore(Instant.now().minusSeconds(60))) {  // Adding a buffer to ensure token validity
+            try {
+                generateJWT(appId, privateKeyPath);
+            } catch (IOException | GeneralSecurityException exception) {
+                log.warn("Could not refresh the JWT due to exception: {}", exception.getMessage());
```
jeetchoudhary, line 93: Wouldn't this cause a problem? If we simply ignore this exception and default the value to false in the `isGithubAppEnabledOnRepository` method, it means that if the JWT refresh fails, we'll end up creating PRs even if the repository isn't onboarded to Renovate.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/PullRequestsTest.java
```diff
@@ -17,148 +17,151 @@
...
-        pullRequests.prepareToCreate(ns, pullRequestSender, contentsFoundWithImage,
-                gitForkBranch, dockerfileGitHubUtil, rateLimiter);
-
-        //Verify that the DFIU PR is skipped for the first repo, but is sent to the other two repos
-        verify(dockerfileGitHubUtil, times(2)).changeDockerfiles(eq(ns),
-                eq(pathToDockerfilesInParentRepo),
-                eq(gitHubContentToProcess), anyList(), eq(gitForkBranch),
-                eq(rateLimiter));
-    }
+//    @Test
```
jeetchoudhary: commented tests?

jeetchoudhary (changes requested):
Q: Are you sure you want to merge this change into the main branch? You can achieve what you need by using a branch artifact instead.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
ghost: But `github` is a GithubBuilder object, not Github though, right?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
AvvariSaiBharadwaj: Yes, but I am trying to be cautious. Since we are dealing with docker patches, isn't it better to get both DFIU and Renovate PRs and have a confusing experience, rather than not receive either PR and fall behind on your patches?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
AvvariSaiBharadwaj: Same comment as above. I am trying to be cautious. Lmk if you think otherwise.

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/PullRequestsTest.java
AvvariSaiBharadwaj: Working on fixing them. Wanted to get the code reviewed beforehand.

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/RenovateUtil.java
AvvariSaiBharadwaj: Pls refer to https://github.com/AvvariSaiBharadwaj/dockerfile-image-update/blob/excludeRenovateRepos/dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java#L45-L48. `github` is generated by `githubBuilder.build()` which returns a github object (https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubBuilder.java#L508)

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
```diff
@@ -112,6 +112,14 @@ static ArgumentParser getArgumentParser() {
                 .setDefault(false) //To prevent null from being returned by the argument
                 .required(false)
                 .help("Enable debug logging, including git wire logs.");
+        parser.addArgument("--" + SKIP_GITHUB_APP_ID)
+                .type(String.class)
```
afalko, line 116: Aren't IDs always integers?

ghost: @afalko It does not matter as it would be parsed as a parameter in a request API call to Github API to generate a JWT token. We have tested this

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
hanelliot-sfdc: @jeetchoudhary Why would we end up creating PRs even if the repository isn't onboarded to Renovate (if they are not onboarded to Renovate then Renovate should not create a PR no matter if the JWT token fails or not right)? You mean even if the repository that is not even onboarded to DFIU?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
afalko: Why not have the parser do the parsing for you upfront here?

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
afalko: Agree here

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
```diff
@@ -0,0 +1,134 @@
...
+    private void generateJWT(String appId, String privateKeyPath) throws IOException, GeneralSecurityException {
+        Security.addProvider(new org.bouncycastle.jce.provider.BouncyCastleProvider());
+        RSAPrivateKey privateKey = getRSAPrivateKey(privateKeyPath);
+
+        Algorithm algorithm = Algorithm.RSA256(null, privateKey);
+        Instant now = Instant.now();
+        jwt = JWT.create()
+                .withIssuer(appId)
+                .withIssuedAt(Date.from(now))
+                .withExpiresAt(Date.from(now.plusSeconds(600))) // 10 minutes expiration
```
afalko, line 115: why is this hardcoded?

afalko: what will the behavior be if you set this really short (for example 1 second)

ghost: @afalko we would like to avoid as much API calls as possible. Setting to 1 second means that we have to make an API call to regenerate a new JWT token every 1 second. Taking into account that the DFIU takes hours to run, I believe it is not efficient

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/CommandLine.java
ghost: @afalko the reason for that is because appID will be parsed into the JWT token generation API call under the `.withIssuer(appId)` subcall, and this has to be parsed as a string type. You can see more here: https://www.baeldung.com/java-auth0-jwt

We can change it to `Integer` type if you truly think it is necessary and purposeful (to comply with the App ID integer type from Github App) and convert it to string when parsed into this API call

afalko: Gotcha - pass-through might be fine, but it likely be a difficult error message to understand (vs. hey, you need an integer here). I think it is purposeful to make this integer for that release, but not necessary :)

avimanyum (dismissed):

dockerfile-image-update/src/test/java/com/salesforce/dockerfileimageupdate/utils/PullRequestsTest.java
ghost: @jeetchoudhary added the unit tests with the change

dockerfile-image-update/src/main/java/com/salesforce/dockerfileimageupdate/utils/GithubAppCheck.java
ghost: > Yes, but I am trying to be cautious. Since we are dealing with docker patches, isn't it better to get both DFIU and Renovate PRs and have a confusing experience, rather than not receive either PR and fall behind on your patches?

I agree with Sai in this case

avimanyum (dismissed):

ghost:
See this issue regarding the `docker compose` change: https://github.com/actions/runner-images/issues/10451

avimanyum (approved):

